PR #89, opened 2014-09-18
Fix modulo bias, comment on unsafe code.

therealmik:
Note that I haven't tested these changes (I did it in a github window) - but I'm sure you get the gist of them :)


special:
I extracted the randomPrintable bias fix in #178. I think the rest of this is unnecessary.

#178 also makes RNG errors fatal.

PR #106, opened 2014-09-28
Updated German translation

ghost:
Some fixes in German translation. I have updated the ricochet_de.ts file with Linguist.


special:
Translation changes should go into [Transifex](https://www.transifex.com/projects/p/ricochet/). If you're not comfortable using it, please get in touch with me (or the language's maintainer) directly.

PR #121, opened 2014-10-22
Give conversation windows sense of time

jan0e:
This is a feature I missed, so here is my approach to include timestamps in chat windows. I have to admit that I like timestamps on every message so I shamelessly implemented that first (it was the easiest, too). But it is disabled by default, it can be activated in Preferences -> General. The default behaviour is:
 - Add a date/time separator whenever a new conversation is started or one hour has passed since the last exchange of messages. This uses the separator which is shown when the chat partner is offline. When a new conversation is started and the partner is offline this text is still shown.
 - I liked the idead of @obvio171 so one can double click a bubble to see the timestamp of the message. This might be a little tricky, because you have to click the bubble not the text in it. This is because if you include the MouseArea on top of the text the text can't be selected anymore (while writing these lines I'm thinking a tool tip on mouse over might be a better solution).

Here are a few examples:

Default behaviour on Xubuntu with locale set to en_US. Bubble 2 got double clicked
screenshot - 10222014 - 01 01 53 am

Default behaviour on Xubuntu with locale set to de_DE. Bubble 2 got double clicked, partner went offline before sending message 2.
screenshot - 22 10 2014 - 10 03 28

Default behaviour on Windows with threshold set to a few seconds.
unbenannt2

Same window with 'show timestamp on every chat message' activated
unbenannt

What do you think?

### DIFF
--- a/src/core/ConversationModel.cpp
+++ b/src/core/ConversationModel.cpp
@@ -122,6 +122,7 @@ QHash<int,QByteArray> ConversationModel::roleNames() const
     roles[IsOutgoingRole] = "isOutgoing";
     roles[StatusRole] = "status";
     roles[SectionRole] = "section";
+    roles[TimespanRole] = "timespan";
     return roles;
 }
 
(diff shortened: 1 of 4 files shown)


special:
I'm going to close this as abandoned - but most if it has been merged in #154. For displaying the time of individual messages, I think we should try doing it in a different way that doesn't dynamically resize delegates.

PR #127, opened 2014-11-07
Update BUILDING.md

AlexTalker:
Add Tor in depends list and instruction for Arch Linux(use AUR package)
Sorry if i've make grammar error, please fix it and tell me!


special:
Closing this, as mentioned above.

I think the right solution here is to get a list of distribution maintainers, verify that the packages are done well, and stay in contact to keep them up to date. Then we can look at mentioning the good ones from the README or other sources.

PR #133, opened 2015-02-24
Fix network setup getting stuck when tor fails

special:
If tor failed before the first page was visible on screen, it would
appear to be stuck on the first setup screen, rather than displaying the
error (often that the executable is missing).

Also disabled the window height animation on Windows, as it causes some
graphical glitches.


rburchell:
LGTM

PR #137, opened 2015-03-06
Minor fixes for Qt 5.4 related issues

special:


src/ui/qml/MainToolBar.qml
```diff
@@ -6,6 +6,8 @@ import im.ricochet 1.0
 ToolBar {
     Layout.minimumWidth: 200
     Layout.fillWidth: true
+    // Necessary to avoid oversized toolbars, e.g. OS X with Qt 5.4.1
+    implicitHeight: toolBarLayout.height + __style.padding.top + __style.padding.bottom
```
rburchell: mmm, private properties. :)

rburchell:
Ok, although maybe it would be an idea to start thinking what can be done to replace the toolbar?

special:
Replacing the toolbar will be design changes, and design changes require more thought. This is just repairing behavior to what it was prior to Qt 5.4.

There will be design changes.

PR #138, opened 2015-03-10
Connection and user handling for new protocol

special:
First section of my [protocol branch](https://github.com/special/ricochet/compare/protocol). This implements the new protocol in ContactUser for inbound and outbound connections.

The next PR will implement chat, then contact requests, then some overall fixups and changes. All of this code is still enabled only when built with `DEFINES+=PROTOCOL_NEW`.

### DIFF
--- a/src/core/ContactUser.cpp
+++ b/src/core/ContactUser.cpp
@@ -223,6 +223,20 @@ void ContactUser::onDisconnected()
     qDebug() << "Contact" << uniqueID << "disconnected";
     m_settings->write("lastConnected", QDateTime::currentDateTime());
 
+#ifdef PROTOCOL_NEW
+    if (m_connection) {
+        if (m_connection->isConnected()) {
+            BUG() << "onDisconnected called, but connection is still connected";
+            return;
+        }
+
+        m_connection->deleteLater();
+        m_connection = 0;
+    } else {
+        BUG() << "onDisconnected called without a connection";
+    }
+#endif
+
     updateStatus();
     emit disconnected();
 }
@@ -398,13 +412,17 @@ void ContactUser::assignConnection(Protocol::Connection *connection)
         m_outgoingSocket = 0;
     }
 
+    if (m_connection && !m_connection->isConnected()) {
+        qDebug() << "Replacing dead connection with new connection";
+        clearConnection();
+    }
+
     /* If the existing connection is in the same direction as the new one,
      * always use the new one.
      */
     if (m_connection && connection->direction() == m_connection->direction()) {
         qDebug() << "Replacing existing connection with contact because the new one goes the same direction";
-        // XXX things
-        m_connection = 0;
+        clearConnection();
     }
 
     /* If the existing connection is more than 30 seconds old, measured from
@@ -412,8 +430,7 @@ void ContactUser::assignConnection(Protocol::Connection *connection)
      */
     if (m_connection && m_connection->age() > 30) {
         qDebug() << "Replacing existing connection with contact because it's more than 30 seconds old";
-        // XXX do things to get rid of the existing connection
-        m_connection = 0;
+        clearConnection();
     }
 
     /* Otherwise, close the connection for which the server's onion-formatted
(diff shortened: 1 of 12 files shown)


src/core/ContactUser.h
```diff
@@ -135,11 +170,18 @@ private slots:
     void onSettingsModified(const QString &key, const QJsonValue &value);
 
 private:
+#ifdef PROTOCOL_NEW
+    // XXX be paranoid about tracking deletion of m_connection
```
special: A commit in a later series addresses this by making m_connection a QPointer, and cleaning up its deletion logic a bit.

src/core/UserIdentity.cpp
```diff
@@ -152,3 +178,75 @@ bool UserIdentity::isServicePublished() const
...
+ *
+ * If the connection successfully completes authentication,
+ * handleIncomingAuthedConnection is called to link it to a ContactUser
+ * (if applicable) and set the purpose.
+ */
+void UserIdentity::onIncomingConnection()
+{
+    while (m_incomingServer->hasPendingConnections()) {
+        QTcpSocket *socket = m_incomingServer->nextPendingConnection();
+        socket->setProperty("localHostname", m_hiddenService->hostname());
```
rburchell: This property appears to be unused?

src/core/ContactUser.cpp
```diff
@@ -307,4 +381,132 @@ void ContactUser::incomingProtocolSocket(QTcpSocket *socket)
...
+     */
+    if (m_connection && connection->direction() == m_connection->direction()) {
+        qDebug() << "Replacing existing connection with contact because the new one goes the same direction";
+        clearConnection();
+    }
+
+    /* If the existing connection is more than 30 seconds old, measured from
+     * when it was successfully established, it's replaced with the new one.
+     */
+    if (m_connection && m_connection->age() > 30) {
```
rburchell: What's the purpose of this?

src/core/ContactUser.cpp
```diff
@@ -307,4 +381,132 @@ void ContactUser::incomingProtocolSocket(QTcpSocket *socket)
...
+void ContactUser::clearConnection()
+{
+    if (!m_connection)
+        return;
+
+    disconnect(m_connection, 0, this, 0);
+    if (m_connection->isConnected()) {
+        connect(m_connection, &Protocol::Connection::closed, m_connection, &QObject::deleteLater);
+        m_connection->close();
+    } else
```
rburchell: } else {
...
}

src/core/ContactUser.cpp
```diff
@@ -179,6 +237,19 @@ void ContactUser::onDisconnected()
...
+#ifdef PROTOCOL_NEW
+    if (m_connection) {
+        if (m_connection->isConnected()) {
+            BUG() << "onDisconnected called, but connection is still connected";
+            return;
+        }
+
+        m_connection->deleteLater();
+        m_connection = 0;
+    } else
```
rburchell: } else {
...
}

src/core/UserIdentity.cpp
special: It is used to implement `Connection::serverHostname`. It could be done more clearly by adding a `Connection::setServerHostname` method. Do you think that is necessary?

src/core/ContactUser.cpp
special: 1. We have an active outbound connection
2. That connection has failed, but we haven't noticed the failure on this end
3. The peer has noticed failure, and makes an inbound connection to repair it
4. The peer loses the `preferOutbound` comparison below

In this case, we would keep the old, actually-dead outbound connection instead of using the new inbound one, until we noticed the failure. This does happen in practice. Always preferring a new connection over a >30 second old connection solves it.

The assumptions behind this are:

1. 30 seconds is long enough to avoid the "race for connection" behavior that `preferOutbound` handles. This is reasonable, because it's twice as long as the authentication timeout.
2. If the peer is making a new connection instead of using the established one, they probably know that it has failed and we don't.

There's not a security risk to letting the peer control this decision (and they already can).

src/core/UserIdentity.cpp
rburchell: It would certainly be a lot more visible what was going on.

special: Ah, but it would make constructing a Connection significantly more annoying: this value must be known at construction time.

PR #139, opened 2015-03-16
Chat message implementation for new protocol

special:


src/protocol/ChatChannel.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    pendingMessages.insert(id);
+    return true;
+}
+
+void ChatChannel::handleChatMessage(const Data::Chat::ChatMessage &message)
+{
+    QScopedPointer<Data::Chat::ChatAcknowledge> response(new Data::Chat::ChatAcknowledge);
+
+    // QString::fromStdString decodes the string as UTF-8, replacing all invalid sequences and
```
rburchell: This is documented as using fromAscii. I know that the functionality *nowdays* will use UTF8 conversion, but it might be smarter to be explicit about this, using fromUtf8

src/protocol/ChatChannel.cpp
```diff
@@ -0,0 +1,196 @@
...
+    // codepoints with the unicode replacement character.
+    QString text = QString::fromStdString(message.message_text());
+
+    if (direction() != Inbound) {
+        qWarning() << "Rejected inbound message on an outbound chat channel";
+        response->set_accepted(false);
+    } else if (text.isEmpty()) {
+        qWarning() << "Rejected empty chat message";
+        response->set_accepted(false);
+    } else if (text.size() > MessageMaxCharacters) {
```
rburchell: Should this be >, or >=?

src/protocol/ChatChannel.cpp
special: The documentation was wrong in 5.3; it's fixed now: https://codereview.qt-project.org/#/c/93509/

I tried to find reasons to use some more complicated decoding process, but I couldn't come up with any.

src/protocol/ChatChannel.cpp
special: >. This is the maximum length of a message. It seems to be used that way consistently.

rburchell:
:+1:

PR #140, opened 2015-03-16
Alert contacts using old protocol of the need to upgrade

special:
Ricochet 1.0.x clients have no way of showing an error when they try to connect to clients running the newer protocol. We decided backwards compatibility would be too much extra work in this particular case, and it's still early enough to make a hard break without inconveniencing people too much.

This is an attempt to make that hard break less hard. When an _outbound_ connection is made to a known contact, and they negotiate the old protocol, it will send a chat message directing them to figure out how to upgrade:

example

That message is sent only once per contact, and only for outbound connections.

I don't like this solution, but since we currently lack a reliable way to tell users to upgrade, this seems like a step worth taking. Thoughts?

### DIFF
--- a/doc/protocol.md
+++ b/doc/protocol.md
@@ -129,6 +129,38 @@ document:
 
 ## Specification
 
+### Introduction and version negotiation
+
+Immediately after establishing a connection, the client side must send an introduction message
+identifying versions of the protocol that it is able to use. The server responds with one of
+those versions, or an error indicating that no compatible version exists.
+
+This step exists to enable smoother protocol changes in the future, and for better compatibility
+with old clients.
+
+The client begins the connection by sending the following raw sequence of bytes:
+
+```
+0x49
+0x4D
+nVersions          // One byte, number of supported protocol versions
+nVersions times:
+    version        // One byte, protocol version number
+```
+
+The total size is 3 plus the number of supported versions bytes. The server side of the connection
+must respond with a single byte for the selected version number, or 0xFF if no suitable version
+is found.
+
+This document describes protocol version 1. Known versions are:
+```
+0                  The Ricochet 1.0 protocol
+1                  This document
+```
+
+If the negotiation is successful, the connection can be immediately used to begin exchanging messages
+(the packet layer, below).
+
 ### Packet layer
 
 The base layer on the connection is a trivial packet structure:
--- a/src/protocol/Connection.cpp
+++ b/src/protocol/Connection.cpp
@@ -54,6 +54,7 @@ ConnectionPrivate::ConnectionPrivate(Connection *qq)
     , direction(Connection::ClientSide)
     , purpose(Connection::Purpose::Unknown)
     , wasClosed(false)
+    , handshakeDone(false)
     , nextOutboundChannelId(-1)
 {
     ageTimer.start();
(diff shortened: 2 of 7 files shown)


rburchell:
I assume you'll put something on that URL. Right now it just redirects to Github.

rburchell:
Alternatively, maybe putting a markdown doc in the source tree (and linking to that) would work?

src/protocol/Connection.cpp
```diff
@@ -222,6 +231,91 @@ void ConnectionPrivate::socketDisconnected()
...
+                emit q->versionNegotiationFailed();
+                socket->abort();
+                return;
+            } else
+                emit q->ready();
+        } else if (direction == Connection::ServerSide && available >= 3) {
+            // Expecting at least 3 bytes
+            uchar intro[3] = { 0 };
+            qint64 re = socket->peek(reinterpret_cast<char*>(intro), sizeof(intro));
+            if (re < (int)sizeof(intro)) {
```
rburchell: Isn't this worthy of BUG()? Since bytesAvailable looked correct

doc/protocol.md
```diff
@@ -129,6 +129,38 @@ document:
...
+
+This step exists to enable smoother protocol changes in the future, and for better compatibility
+with old clients.
+
+The client begins the connection by sending the following raw sequence of bytes:
+
+```
+0x49
+0x4D
+nVersions          // One byte, number of supported protocol versions
```
rburchell: Not that I expect you'll have a billion versions, but a single byte is a pretty big restriction for something sent only once. You could increase it a bit.

rburchell: I also think the nVersions thing is a bit strange. Why not just have a serially increasing protocol version, and have the negotiation mechanism be MIN(myVersion, remoteVersion)?

rburchell: Though, I suppose that allows unpublishing a version retroactively should that be required.

src/protocol/OutboundConnector.cpp
```diff
@@ -256,6 +257,25 @@ void OutboundConnectorPrivate::onConnected()
     socket->setReconnectEnabled(false);
     socket = 0;
 
+    connect(connection, &Connection::ready, this, &OutboundConnectorPrivate::startAuthentication);
+    // XXX Needs special treatment in UI (along with some other error types here)
```
rburchell: Just don't leave this one undone forever, or you may end up regretting it ;)

doc/protocol.md
special: This is identical to the 1.0 protocol's version negotiation; being compatible with that is useful.

It's one byte because that keeps byte-fiddling to a minimum (no endianness, etc). Not using a minimum means we can remove support for a version and have this recognized to display a useful message.

special:
Iterating on the message text:

Original:
> [automatic message] I'm using a newer version of Ricochet, which is not compatible with yours. This is a one-time change to help improve Ricochet in the future. See https://ricochet.im/upgrade for instructions.

Revised:
> [automatic message] I'm using a newer version of Ricochet that cannot connect to yours anymore. This is a one-time change to help improve Ricochet. See https://ricochet.im/upgrade for instructions on how to update.

Patrick's proposal:
> [automatic message] I'm using a newer version of Ricochet that is not compatible with yours. This is a one-time change to help improve Ricochet. See https://ricochet.im/upgrade for instructions on downloading and installing the latest version of Ricochet. Once you have upgraded I will be able to see your messages again, and you mine.

rburchell:
I like the addition about explaining explicitly that upgrading will fix things.

src/protocol/Connection.cpp
special: I think no: I assume there are all sorts of reasons for a read/peek to fail. I'm mostly only concerned if peek succeeds and then read fails, or if we get some data but less than bytesAvailable promised.

special:
>    [automatic message] I'm using a newer version of Ricochet that is not compatible with yours. This is a one-time change to help improve Ricochet. See https://ricochet.im/upgrade for instructions on getting the latest version. Once you have upgraded, I will be able to see your messages again.

special:
The XXX about UI for protocol negotiation errors will be addressed in a separate PR for easier review.

PR #141, opened 2015-03-17
Remove the hidden service self-testing step

### LINKED ISSUE #26: Remove the unnecessary and confusing hidden service self test
Specs: Torsion v1.0.0-10-g2c52502 on Debian 7 (64bit) in VirtualBox:

Repro Steps:
1) Open the Torsion binary and successfully connect to the Tor network.
2) Click the "Preferences" button/icon.
3) Click the "Tor" tab.
4) Wait 3 minutes looking for a notice about "Closing stream for '[scrubbed].onion': hidden service is unavailable (try again later)."

Results: 
The Torsion program is making onion connections when the user has no contacts. This is worrying.

Expected Result: The Torsion program only make authorized onion connection attempts. 

Notes:
Perhaps these unexplained and failing onion attempts are expected behavior as Torison could be using a Hidden service for updates or for some other reason, but I don't recall any comment in the design or protocol documents mentioning it. Could it be attempting to connect to my own published ID (similar to how Torchat does it by default)?

special:
It's not necessary to self-test connectivity of the hidden service
immediately after we publish it. This information isn't used for
anything valuable, and in every observed situation, we already have a
good idea of whether our service is connectable or not based on tor's
circuit status.

It has also raised some concerns, because a new Ricochet client with no
contacts will start trying to connect to some scrubbed service in the
logs.

Fixes #26

### DIFF
--- a/src/core/UserIdentity.cpp
+++ b/src/core/UserIdentity.cpp
@@ -173,11 +173,6 @@ bool UserIdentity::isServiceOnline() const
     return m_hiddenService && m_hiddenService->status() == Tor::HiddenService::Online;
 }
 
-bool UserIdentity::isServicePublished() const
-{
-    return m_hiddenService && m_hiddenService->status() >= Tor::HiddenService::Published;
-}
-
 #ifdef PROTOCOL_NEW
 /* Handle an incoming connection to this service
  *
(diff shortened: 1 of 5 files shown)


rburchell:
:+1:

PR #144, opened 2015-03-19
Contact requests in new protocol

special:
This one is a bit bigger, and complicated in parts.

### DIFF
--- a/src/core/IncomingRequestManager.cpp
+++ b/src/core/IncomingRequestManager.cpp
@@ -34,6 +34,7 @@
 #include "IncomingRequestManager.h"
 #include "ContactsManager.h"
 #include "OutgoingContactRequest.h"
+#include "ContactIDValidator.h"
 #include "protocol/ContactRequestServer.h"
 #include <QDebug>
 
@@ -214,6 +215,11 @@ void IncomingContactRequest::removeRequest()
 {
     SettingsObject settings(QStringLiteral("contactRequests.%1").arg(QLatin1String(m_hostname)));
     settings.undefine();
+}
+
+QString IncomingContactRequest::contactId() const
+{
+    return ContactIDValidator::idFromHostname(hostname());
 }
 
 void IncomingContactRequest::setRemoteSecret(const QByteArray &remoteSecret)
--- a/src/core/IncomingRequestManager.cpp
+++ b/src/core/IncomingRequestManager.cpp
@@ -157,16 +157,11 @@ void IncomingRequestManager::requestReceived()
         request->setChannel(channel);
     }
 
-    // Check if this request matches any existing users, including any outgoing requests
-    ContactUser *existingUser = contacts->lookupHostname(hostname);
-    if (existingUser) {
-        // Implicitly accept a matching outgoing request
-        if (existingUser->contactRequest())
-            existingUser->contactRequest()->accept();
-
-        // Implicitly accept this request
-        // XXX Test this case to make sure response status gets handled correctly
-        request->accept(existingUser);
+    /* It shouldn't be possible to get an incoming contact request for a known
+     * contact, including an outgoing request. Those are implicitly accepted at
+     * a different level. */
+    if (contacts->lookupHostname(hostname)) {
+        BUG() << "Created an inbound contact request matching a known contact; this shouldn't be allowed";
         return;
     }
 
--- a/src/core/IncomingRequestManager.cpp
+++ b/src/core/IncomingRequestManager.cpp
@@ -49,9 +49,17 @@ void IncomingRequestManager::loadRequests()
 {
     SettingsObject settings(QStringLiteral("contactRequests"));
 
-    foreach (const QString &host, settings.data().keys())
-    {
-        IncomingContactRequest *request = new IncomingContactRequest(this, host.toLatin1());
+    foreach (const QString &hostStr, settings.data().keys()) {
+        QByteArray host = hostStr.toLatin1();
+#ifdef PROTOCOL_NEW
+        if (!host.endsWith(".onion"))
+            host.append(".onion");
+#else
+        if (host.endsWith(".onion"))
+            host.chop(6);
+#endif
+
+        IncomingContactRequest *request = new IncomingContactRequest(this, host);
         request->load();
 
         m_requests.append(request);
@@ -70,7 +78,12 @@ QList<QObject*> IncomingRequestManager::requestObjects() const
 
 IncomingContactRequest *IncomingRequestManager::requestFromHostname(const QByteArray &hostname)
 {
+#ifdef PROTOCOL_NEW
+    Q_ASSERT(hostname.endsWith(".onion"));
+#else
     Q_ASSERT(!hostname.endsWith(".onion"));
+#endif
+
     Q_ASSERT(hostname == hostname.toLower());
 
     for (QList<IncomingContactRequest*>::ConstIterator it = m_requests.begin(); it != m_requests.end(); ++it)
@@ -174,14 +187,27 @@ IncomingContactRequest::IncomingContactRequest(IncomingRequestManager *m, const 
     : QObject(m), manager(m), connection(c), m_hostname(h)
 {
     Q_ASSERT(manager);
+#ifdef PROTOCOL_NEW
+    Q_ASSERT(m_hostname.endsWith(".onion"));
+#else
     Q_ASSERT(m_hostname.size() == 16);
+#endif
 
     qDebug() << "Created contact request from" << m_hostname << (connection ? "with" : "without") << "connection";
 }
 
+QString IncomingContactRequest::settingsKey() const
+{
+    QString key = QString(QLatin1String(m_hostname));
+#ifdef PROTOCOL_NEW
+    key.chop(QStringLiteral(".onion").size());
+#endif
+    return QStringLiteral("contactRequests.%1").arg(key);
+}
+
 void IncomingContactRequest::load()
 {
-    SettingsObject settings(QStringLiteral("contactRequests.%1").arg(QLatin1String(m_hostname)));
+    SettingsObject settings(settingsKey());
 
     setRemoteSecret(settings.read<Base64Encode>("remoteSecret"));
     setNickname(settings.read("nickname").toString());
@@ -193,7 +219,7 @@ void IncomingContactRequest::load()
 
 void IncomingContactRequest::save()
 {
-    SettingsObject settings(QStringLiteral("contactRequests.%1").arg(QLatin1String(m_hostname)));
+    SettingsObject settings(settingsKey());
 
     settings.write("remoteSecret", Base64Encode(remoteSecret()));
     settings.write("nickname", nickname());
@@ -213,8 +239,7 @@ void IncomingContactRequest::renew()
 
 void IncomingContactRequest::removeRequest()
 {
-    SettingsObject settings(QStringLiteral("contactRequests.%1").arg(QLatin1String(m_hostname)));
-    settings.undefine();
+    SettingsObject(settingsKey()).undefine();
 }
 
 QString IncomingContactRequest::contactId() const
--- a/src/core/IncomingRequestManager.cpp
+++ b/src/core/IncomingRequestManager.cpp
@@ -35,14 +35,35 @@
 #include "ContactsManager.h"
 #include "OutgoingContactRequest.h"
 #include "ContactIDValidator.h"
+#include "utils/Useful.h"
+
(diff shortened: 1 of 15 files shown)


src/core/IncomingRequestManager.cpp
```diff
@@ -34,22 +34,54 @@
...
+    );
+#endif
 }
 
 void IncomingRequestManager::loadRequests()
 {
     SettingsObject settings(QStringLiteral("contactRequests"));
 
-    foreach (const QString &host, settings.data().keys())
+    foreach (QString host, settings.data().keys())
```
rburchell: Why the const& removal?

src/core/IncomingRequestManager.cpp
```diff
@@ -34,22 +34,54 @@
...
     {
+#ifdef PROTOCOL_NEW
+        if (!host.endsWith(QStringLiteral(".onion")))
```
rburchell: Don't use multiple QStringLiteral. Use an inline function, otherwise, the string will be duplicated multiple times in the binary, bloating the size of it.

For that matter, use QLatin1String where you can - as then you don't need the QString construction for the argument at all.

src/core/IncomingRequestManager.cpp
```diff
@@ -168,21 +284,44 @@ bool IncomingRequestManager::isHostnameRejected(const QByteArray &hostname) cons
...
+    : QObject(m)
+    , manager(m)
+#ifndef PROTOCOL_NEW
+    , connection(c)
+#endif
+    , m_hostname(h)
 {
     Q_ASSERT(manager);
+#ifdef PROTOCOL_NEW
+    Q_ASSERT(m_hostname.endsWith(".onion"));
```
rburchell: QLatin1String

src/core/IncomingRequestManager.cpp
```diff
@@ -192,9 +331,11 @@ void IncomingContactRequest::load()
 
 void IncomingContactRequest::save()
 {
-    SettingsObject settings(QStringLiteral("contactRequests.%1").arg(QLatin1String(m_hostname)));
+    SettingsObject settings(settingsKey());
 
+#ifndef PROTOCOL_NEW
     settings.write("remoteSecret", Base64Encode(remoteSecret()));
```
rburchell: Question: how are contact request, saved from old protocol, going to work when upgrading?

src/core/IncomingRequestManager.cpp
```diff
@@ -233,6 +378,51 @@ void IncomingContactRequest::setNickname(const QString &nickname)
...
+            {
+                qDebug() << "Closing connection attached to an IncomingContactRequest because ContactRequestChannel was closed";
+                connection->close();
+            }
+        }
+    );
+
+    qDebug() << "Assigning connection to IncomingContactRequest from" << m_hostname;
+    connection = channel->connection();
+    connection->setParent(this);
```
rburchell: The channel stays parented to the manager, but the connection inside the channel is reparented. What's the ownership semantics supposed to be here?

src/core/IncomingRequestManager.cpp
```diff
@@ -233,6 +378,51 @@ void IncomingContactRequest::setNickname(const QString &nickname)
...
+    if (!connection->setPurpose(Protocol::Connection::Purpose::InboundRequest)) {
+        qDebug() << "Setting purpose on incoming contact request connection failed; killing connection";
```
rburchell: BUG() worthy? (the other setPurpose calls in this diff do a qWarning, btw; should they be BUG() too, or should this be qWarning then?)

src/core/IncomingRequestManager.cpp
```diff
@@ -233,6 +378,51 @@ void IncomingContactRequest::setNickname(const QString &nickname)
...
+        connection->close();
+        connection = 0;
+        return;
```
rburchell: What is responsible for deleting connection, given it's parented to us already?

src/core/IncomingRequestManager.cpp
```diff
@@ -233,6 +378,51 @@ void IncomingContactRequest::setNickname(const QString &nickname)
...
+    connect(connection.data(), &Protocol::Connection::closed, this,
+        [this]() {
+            if (connection && !connection->isConnected()) {
+                connection->deleteLater();
+                connection.clear();
+            }
+        }
+    );
+
+    emit hasActiveConnectionChanged();
```
rburchell: Is the ordering right here? Should activeConnectionChanged change after setting properties on it which the signal handler may want to access?

src/core/IncomingRequestManager.cpp
```diff
@@ -233,6 +378,51 @@ void IncomingContactRequest::setNickname(const QString &nickname)
...
+    connection = channel->connection();
+    connection->setParent(this);
+    if (!connection->setPurpose(Protocol::Connection::Purpose::InboundRequest)) {
+        qDebug() << "Setting purpose on incoming contact request connection failed; killing connection";
+        connection->close();
+        connection = 0;
+        return;
+    }
+
+    connect(connection.data(), &Protocol::Connection::closed, this,
```
rburchell: You don't disconnect if setChannel is called with a different channel.

src/core/IncomingRequestManager.cpp
```diff
@@ -246,29 +436,49 @@ void IncomingContactRequest::setConnection(ContactRequestServer *c)
...
+    }
+#else
     user->settings()->write("remoteSecret", Base64Encode(remoteSecret()));
 
-    /* If there is a connection, send the accept message and morph it to a primary connection */
-    if (connection)
-    {
+    if (connection) {
         connection.data()->sendAccept(user);
         connection = (ContactRequestServer*)0;
```
rburchell: If you're using C++11 (which you seem to be doing), then you could start using nullptr in some places where it makes sense.

src/core/IncomingRequestManager.cpp
special: Because host is modified based on the code block below, before it's used to the IncomingContactRequest constructor

src/core/OutgoingContactRequest.cpp
```diff
@@ -107,7 +119,11 @@ void OutgoingContactRequest::attemptAutoAccept()
 {
     /* Check if there is an existing incoming request that matches this one; if so, treat this as accepted
      * automatically and accept that incoming request for this user */
+#ifdef PROTOCOL_NEW
+    QByteArray hostname = user->hostname().toLatin1();
+#else
     QByteArray hostname = user->hostname().left(16).toLatin1();
```
rburchell: I'm happy to see that magic number going away.

src/core/IncomingRequestManager.cpp
special: True. It actually looks like this should be a QByteArray instead, because it's just `toLatin1`'d below.

src/core/IncomingRequestManager.cpp
special: The secret used to be required, and is no longer required; it's all based on keys now. So, no problems.

src/protocol/Connection.h
```diff
@@ -125,7 +125,9 @@ class Connection : public QObject
      */
     enum class Purpose {
         Unknown,
```
rburchell: It seems a little weird to have unknown as the first value of the enum. But not important.

src/protocol/ContactRequestChannel.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if (!request->HasExtension(Data::ContactRequest::contact_request)) {
+        result->set_error_message(QStringLiteral("Expected a request object").toStdString());
+        return false;
+    }
+
+    ContactRequest contactData = request->GetExtension(Data::ContactRequest::contact_request);
+    QString nickname = QString::fromStdString(contactData.nickname());
+    QString message = QString::fromStdString(contactData.message_text());
+
+    if (message.size() > Data::ContactRequest::MessageMaxCharacters ||
```
rburchell: This is a recurring comment now, but I don't like how your things are called "MaxSomething" when they're really "MaxSomething+1" due to using > and not >=. So long as it's consistent, I suppose it's fine, but I see it as being something that someone new to the codebase could potentially trip up on.

src/protocol/ContactRequestChannel.cpp
```diff
@@ -0,0 +1,305 @@
...
+        !isAcceptableNickname(nickname))
+    {
+        qDebug() << "Rejecting incoming contact request with invalid nickname/message";
```
rburchell: qWarning

src/protocol/ContactRequestChannel.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    Data::ContactRequest::Response response = result->GetExtension(Data::ContactRequest::response);
+    return handleResponse(&response);
+}
+
+void ContactRequestChannel::receivePacket(const QByteArray &packet)
+{
+    Data::ContactRequest::Response response;
+    if (!response.ParseFromArray(packet.constData(), packet.size())) {
+        closeChannel();
```
rburchell: Does this deserve a qDebug, at least?

src/protocol/ContactRequestChannel.cpp
```diff
@@ -0,0 +1,305 @@
...
+        return;
+    }
+
+    if (!handleResponse(&response))
+        closeChannel();
+}
+
+bool ContactRequestChannel::handleResponse(const Data::ContactRequest::Response *response)
```
rburchell: Some of these debugs look more like BUG()/warnings.

src/protocol/ContactRequestChannel.cpp
```diff
@@ -0,0 +1,305 @@
...
+    if (m_responseStatus > Response::Pending) {
+        qDebug() << "Received a response" << response->status() << "to a contact request which already had a final response" << m_responseStatus;
+        return false;
+    }
+
+    QString message(QString::fromStdString(response->error_message()));
+    if (message.size() > MessageMaxCharacters) {
+        qDebug() << "Received a contact request response with an overly long message of " << message.size() << "characters";
+        message.clear();
+    }
```
rburchell: Yet, you accept it after clearing the message? Intentionally?

src/core/IncomingRequestManager.cpp
rburchell: ... I have no idea how I managed to miss that. True.

rburchell:
Having timeouts as a "XXX" seems a little risky. I take it you're planning on addressing that before switching to new protocol?

src/core/IncomingRequestManager.cpp
special: Good question! I've added the following:

```diff
+    /* Inbound requests are only valid on connections with an Unknown purpose, meaning
+     * they also haven't been claimed by any parent object other than the default. We're
+     * attaching this channel to the request, so we take ownership of the connection here
+     * and set its purpose to InboundRequest. That implicitly means that the channel is
+     * ours too - channels are always owned by the connection.
+     */
```

src/core/IncomingRequestManager.cpp
special: setPurpose itself calls BUG() for invalid transitions, so I think these should all be qWarning at most.

We really need a logging overhaul at some point too. Fortunately most of it is disabled for release builds.

src/core/IncomingRequestManager.cpp
special: Bug! Fixed to not set the parent until after the `closed` handler is added.

src/protocol/ContactRequestChannel.cpp
special: It's not. This says that the message is invalid if there are more than MessageMaxCharacters. So the maximum valid number of characters is the value of MessageMaxCharacters. That fits with my expectation from the word 'max'.

src/protocol/ContactRequestChannel.cpp
special: Ah, a tricky trap!

The `handle*` functions are handling remote data. If we did `BUG()` on them, those 'bug's would be remotely triggerable. That's not dangerous for a release build, but it's not correct either.

src/protocol/ContactRequestChannel.cpp
special: It's a response error message; we can't refuse the response without making everyone confused.

I think this error message is removed in a later set of patches, in favor of an enum.

src/core/IncomingRequestManager.cpp
special: It's a QByteArray at this point

PR #148, opened 2015-03-26
Finishing touches for protocol

special:
A few separate protocol related changes, see individual commits for details.


rburchell:
Looks good.

PR #151, opened 2015-03-31
Show old protocol version contacts as 'Outdated'

special:
This puts contacts that are known to be running the 1.0.x versions in an "Outdated" section, until they upgrade.

One concern is that this string isn't translated, and I doubt we could get many translations for it in time.

### DIFF
--- a/src/core/ContactUser.cpp
+++ b/src/core/ContactUser.cpp
@@ -125,14 +125,18 @@ void ContactUser::updateStatus()
         }
     } else if (settings()->read("rejected").toBool()) {
         newStatus = RequestRejected;
-    } else {
 #ifdef PROTOCOL_NEW
-        newStatus = m_connection && m_connection->isConnected() ? Online : Offline;
+    } else if (m_connection && m_connection->isConnected()) {
+        newStatus = Online;
+    } else if (settings()->read("sentUpgradeNotification").toBool()) {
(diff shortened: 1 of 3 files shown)


src/core/ContactUser.cpp
```diff
@@ -125,15 +125,19 @@ void ContactUser::updateStatus()
         }
     } else if (settings()->read("rejected").toBool()) {
         newStatus = RequestRejected;
-    } else {
 #ifdef PROTOCOL_NEW
-        newStatus = m_connection && m_connection->isConnected() ? Online : Offline;
+    } else if (m_connection && m_connection->isConnected()) {
+        newStatus = Online;
+    } else if (settings()->read("sentUpgradeNotification").toBool()) {
+        newStatus = Outdated;
```
rburchell: What happens once this person then subsequently upgrades?

special: They will connect, and will be online. This flag is unset in onConnected below.

rburchell:
So it is!

:+1:

rburchell:
I don't think string translations are too big a problem, although it is suboptimal. You'll almost certainly be making subsequent releases, and right now (at the early adopter phase, plus a fairly technical userbase), it's probably the case that most of your users have a command of English.

PR #154, opened 2015-03-31
ui: Show a conversation separator if a sufficiently long timespan passes between messages.

rburchell:
When talking with one person throughout multiple conversations over a long
period of time, it can become hard to get a sense of how much time has passed
since a message was received. To help with this, show a separator if a long time
has passed since the last message. For now, a democratically selected hour long
duration has been chosen.

Changes from Jan's original patch:

* Use blocks rather than writing large conditionals inline
* Use Qt's default date format (justification: if the date formatting looks
  "wrong" to someone, it is better that this problem be fixed within Qt, so that
  more places/software benefits)
* Use seconds instead of msec for timespan durations

Done-by: Jan Noertemann <jan.noertemann@uni-dortmund.de>

### DIFF
--- a/src/core/ConversationModel.cpp
+++ b/src/core/ConversationModel.cpp
@@ -323,6 +323,7 @@ QHash<int,QByteArray> ConversationModel::roleNames() const
     roles[IsOutgoingRole] = "isOutgoing";
     roles[StatusRole] = "status";
     roles[SectionRole] = "section";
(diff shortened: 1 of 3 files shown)


rburchell:
This was split out of PR #121.

special:
Looks great. Thanks to both of you

PR #157, opened 2015-04-04
Refresh contact list UI design

### LINKED ISSUE #38: Improving contact list UI
WIP concept to improve on the contact list UI:
torsion

The indicators on top are presence controls. I'm not sure there's much value in a global "offline" presence. I'm not fond of the icons or the behavior for the add contact dialog and preferences.

I have ideas bouncing around about creating contact groups, publishing a hidden service per group, and moving the presence controls onto each individual group. That would allow some separation, selectively controlling online status for different groups of contacts, and might help with a few other design questions. It's not a solid enough idea to turn into a real suggestion yet.

special:
pretty-contacts

Fixes #38.

### DIFF
--- a/src/ui/qml/ContactList.qml
+++ b/src/ui/qml/ContactList.qml
@@ -52,21 +52,46 @@ ScrollView {
         ]
 
         section.property: "status"
-        section.delegate: Label {
-            height: implicitHeight + 8
-            width: parent.width
-            verticalAlignment: Qt.AlignVCenter
-            horizontalAlignment: Qt.AlignHCenter
-            text: {
-                switch (parseInt(section)) {
-                    case ContactUser.Online: return qsTr("Online")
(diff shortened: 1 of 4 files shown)


src/ui/qml/ContactList.qml
```diff
@@ -43,22 +43,55 @@ ScrollView {
...
+
+                font.pointSize: styleHelper.pointSize
+                font.bold: true
+                font.capitalization: Font.SmallCaps
+                color: "#3f454a"
+
+                text: {
+                    // Translation strings are uppercase for legacy reasons, and because they
+                    // should correctly be capitalized. We go lowercase only because it looks
+                    // nicer when using SmallCaps, and that's a display detail.
```
special: I did test that this works correctly for non-latin capitalizations, at least with all of the translations we have.

rburchell:
:+1:

PR #162, opened 2015-04-12
Unmark 'rejected' contacts if they successfully connect

special:
The 'rejected' mark comes from contacts that reject our authentication,
normally because they removed us as a contact. If that person then sends
a new contact request, it will be automatically accepted, and allowed to
connect. In this case, the rejected mark should be removed.

Also added a BUG for the case where a connection is established with a
contact, and the contact's status doesn't match.

### DIFF
--- a/src/core/ContactUser.cpp
+++ b/src/core/ContactUser.cpp
@@ -101,10 +101,10 @@ void ContactUser::updateStatus()
         } else {
             newStatus = RequestPending;
         }
-    } else if (settings()->read("rejected").toBool()) {
-        newStatus = RequestRejected;
     } else if (m_connection && m_connection->isConnected()) {
         newStatus = Online;
+    } else if (settings()->read("rejected").toBool()) {
+        newStatus = RequestRejected;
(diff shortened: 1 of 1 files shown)


rburchell:
:+1:

src/core/ContactUser.cpp
```diff
@@ -219,9 +219,24 @@ void ContactUser::onConnected()
...
         m_settings->write("sentUpgradeNotification", QJsonValue());
 
+    /* The 'rejected' mark comes from failed authentication to someone who we thought was a known
+     * contact. Normally, it would mean that you were removed from that person's contacts. It's
+     * possible for this to be undone; for example, if that person sends you a new contact request,
+     * it will be automatically accepted. If this happens, unset the 'rejected' flag for correct UI.
+     */
+    if (m_settings->read("rejected").toBool()) {
+        qDebug() << "Contact had marked us as rejected, but now they've connected again. Re-enabling.";
+        m_settings->write("rejected", QJsonValue());
```
rburchell: Idle thought: An "unset" method might be clearer than writing null values?

PR #163, opened 2015-04-13
ui: Insert linebreak at cursor position on alt+enter

special:
This linebreak should be inserted at the cursor position, not always at
the end of the string. Also, allow alt+enter in addition to shift+enter
as a way to insert a linebreak.

Suggested by @RedLetterSec


rburchell:
One could split hairs about this being two seperate commits. I'll let you get away with it though ;-)

:+1:

PR #171, opened 2015-04-15
ui: Switch to an icon font for UI icons

special:
Icon fonts handle scaling, positioning, and high DPI much better than
image icons do, and can be considerably smaller. No attempt is made
here to improve on the current UI icons, but one should be.

The font is generated by Fontello, and can be reproduced and modified
using the included .json file.

I haven't tested this on Windows. It shouldn't be merged until that happens.


icons/LICENSE
```diff
@@ -5,3 +5,41 @@ This work is licensed under the Creative Commons Attribution-ShareAlike 4.0
...
 http://creativecommons.org/licenses/by-sa/4.0/ or send a letter to
 Creative Commons, PO Box 1866, Mountain View, CA 94042, USA.
+
+
+## Font Awesome
+
+   Copyright (C) 2012 by Dave Gandy
+
+   Author:    Dave Gandy
+   License:   SIL ()
```
rburchell: Missing the URL present on the other instances.

src/ui/qml/MainToolBar.qml
```diff
@@ -63,9 +79,9 @@ ToolBar {
 
         ToolButton {
             action: preferencesAction
-            iconSource: "qrc:/ui/icons/gear.png"
             implicitHeight: 24
-            implicitWidth: 24
+            style: iconButtonStyle
+            text: "\ue803"
```
rburchell: a "// this is X" at the end might not go astray on these, for those of us that aren't unicode-brained. sure, it's somewhat obvious if you read the context, but for scanning, it helps.

rburchell:
Only real concern is about the font licensing. I'm not familiar with that license (and while on holiday, not able to research).

icons/LICENSE
special: This list is autogenerated by the nifty tool that creates a custom minimal font.

special:
License is very permissive; in this case, I believe it doesn't restrict redistribution or even require attribution. Including the license text here is sufficient.

See also, https://github.com/fontello/fontello/wiki/What-about-license%3F

special:
Tested and works on Windows too.

rburchell:
:+1:

Issue #174, opened 2015-04-19
Build instructions for Debian


hillu:
The `qtdeclarative5-controls-plugin` does not exist in Debian. I have found that

    apt-get install qml-module-qtquick-controls qml-module-qtquick-dialogs

installs the modules necessary to run the program. (At least I get to see the GUI that way.)

It would also be helpful if `ricochet` exited with an error code if it can't initialize the modules. Instead I just got the following messages and the program keeps running:

    QQmlApplicationEngine failed to load component
    qrc:/ui/main.qml:1 module "QtQuick" is not installed
    qrc:/ui/main.qml:2 module "QtQuick.Controls" is not installed
    qrc:/ui/main.qml:3 module "QtQuick.Layouts" is not installed
    qrc:/ui/main.qml:4 module "QtQuick.Window" is not installed
    qrc:/ui/main.qml:1 module "QtQuick" is not installed
    qrc:/ui/main.qml:2 module "QtQuick.Controls" is not installed
    qrc:/ui/main.qml:3 module "QtQuick.Layouts" is not installed
    qrc:/ui/main.qml:4 module "QtQuick.Window" is not installed
    qrc:/ui/main.qml:1 module "QtQuick" is not installed
    qrc:/ui/main.qml:2 module "QtQuick.Controls" is not installed
    qrc:/ui/main.qml:3 module "QtQuick.Layouts" is not installed
    qrc:/ui/main.qml:4 module "QtQuick.Window" is not installed
    qrc:/ui/main.qml:1 module "QtQuick" is not installed
    qrc:/ui/main.qml:2 module "QtQuick.Controls" is not installed
    qrc:/ui/main.qml:3 module "QtQuick.Layouts" is not installed
    qrc:/ui/main.qml:4 module "QtQuick.Window" is not installed

PR #176, opened 2015-04-19
ui: Show an error and exit if the root QML fails to load.

special:
It doesn't seem worth translating and simplifying this error; it should
generally only apply for technical users with their own Qt.

Related to #174.


rburchell:
:+1:

PR #178, opened 2015-04-19
Fix SecureRNG return value and make RNG errors fatal

### LINKED ISSUE #164: Incorrect handling of return value in SecureRNG
Twitter's @mik235 points out:

> https://github.com/ricochet-im/ricochet/blob/master/src/utils/SecureRNG.cpp#L99 that should be "if (r <= 0)"
>
> - https://twitter.com/mik235/status/587440638460690432

Worst case, that could result in a bad challenge secret and authentication issues _if_ openssl fails. Ricochet never generates long-term keys, so there's no impact on that front.

It's also been pointed out in the past that we should fail harder if the RNG does, to be absolutely sure that nothing continues to try to use those values.

special:
Fixes #164, and one of the suggestions from #89.

### DIFF
--- a/src/utils/SecureRNG.cpp
+++ b/src/utils/SecureRNG.cpp
@@ -93,26 +93,17 @@ bool SecureRNG::seed()
     return true;
 }
 
-bool SecureRNG::random(char *buf, int size)
+void SecureRNG::random(char *buf, int size)
 {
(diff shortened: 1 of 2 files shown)


rburchell:
:+1:

Issue #179, opened 2015-04-23
Links have confusing Copy behaviour


gabedwrds:
The context menu for a link in chat has three options: Copy Link, Open with Browser, and Copy, the last one being non-specific to links. It copies selected text, or the entire message if nothing is selected.

If I brought up the context menu by *right*-clicking a link, "Copy" will copy the entire message. This makes sense, if I'm going to have multiple options for what to copy.

If I brought up the context menu by *left*-clicking a link, "Copy" will copy only the link, just like "Copy Link" (presumably it's selected implicitly somehow when activated). This is inconsistent and makes much less sense.

special:
Neat. Thanks for the attention to detail.

I think the options here are:

 * Don't show "Copy" on a link - it's slightly ambiguous, and probably not what was intended
 * Fix "Copy" on left click to behave like it does for right click (copying all text)

I'm not sure which of these I prefer yet. On a vaguely related note, you don't get the link options if you click slightly above or below the link (but still within the bubble). That's slightly annoying.

PR #182, opened 2015-04-26
Separate Copy options to clarify message context menu

gabedwrds:
Add distinct "Copy Message" and "Copy Selection" options in place of the ambiguous "Copy" option

"Copy Message" is available when no text has been selected, or when only a link is selected, such as when a link is left-clicked. Otherwise, "Copy Selection" is available.

Resolves issue #179.

### DIFF
--- a/src/ui/qml/MessageDelegate.qml
+++ b/src/ui/qml/MessageDelegate.qml
@@ -105,7 +105,10 @@ Column {
             selectByMouse: true
             text: LinkedText.parsed(model.text)
 
-            onLinkActivated: delegate.showContextMenu(link)
+            onLinkActivated: {
+                textField.deselect()
+                delegate.showContextMenu(link)
+            }
 
             // Workaround an incomplete fix for QTBUG-31646
             Component.onCompleted: {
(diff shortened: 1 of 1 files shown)


src/ui/qml/MessageDelegate.qml
```diff
@@ -158,18 +158,20 @@ Column {
...
-                        textField.deselect()
-                    }
+                    textField.selectAll()
+                    textField.copy()
+                    textField.deselect()
                 }
             }
+            MenuItem {
+                text: qsTr("Copy Selection")
+                visible: textField.selectedText.length > 0 && textField.selectedText != hoveredLink
```
special: This comparison isn't always working (Qt 5.4, Mac). When mousing over the text `www.google.com`, hoveredLink is actually `http://www.google.com`.

special:
It looks like there are some quirks with selection after left-clicking on a link. The text remains somehow invisibly selected, so if you do:

1. Left click on a link, and close menu
2. Right click on another part of the text

It will still have the link text in selection, and offer "Copy Selection". We probably need to prevent that selection behavior when we're stealing the left click for the context menu.

special:
I also found an unrelated bug while looking at this; the context menu isn't being destroyed properly. Neat!

gabedwrds:
The whole thing would probably be cleaner if links weren't implicitly selected when left-clicked. Let me play with that and see if anything breaks.

src/ui/qml/MessageDelegate.qml
gabedwrds: Shouldn't be an issue anymore since this comparison is no longer necessary.

PR #183, opened 2015-04-26
Fix leak of MessageDelegate context menu objects

special:
This needs a private API of QtQuickControls; I can't find a way to do it
with public API. The alternative is to move the menu instance outside of
MessageDelegate, and have a single instance of it always created for the
view.


rburchell:
Can you at least add a comment noting the private dep? Because I guess when the move to controls 2 happens, you'll need to identify these sort of things easily.

PR #188, opened 2015-05-02
added audio notification support

qsodev:
* contact online notification
 * message received notification

AudioPlayer implementation with QSoundEffect or QMediaPlayer depending on #define


src/core/ContactsManager.h
```diff
@@ -83,9 +83,11 @@ class ContactsManager : public QObject
 
     void unreadCountChanged(ContactUser *user, int unreadCount);
 
+    void contactStatusChanged(ContactUser* user, int status);
+
 private slots:
     void contactDeleted(ContactUser *user);
-    void onUnreadCountChanged();
+    void onUnreadCountChanged();
```
special: Note-to-self: we should fix the whitespace errors before merging.

src/ui/qml/GeneralPreferences.qml
```diff
@@ -24,6 +24,31 @@ ColumnLayout {
...
 
+    CheckBox {
+        text: qsTr("Play audio notifications")
+        checked: uiSettings.data.playAudioNotification || false
+        onCheckedChanged: {
+            uiSettings.write("playAudioNotification", checked)
+        }
+    }
+    RowLayout {
+        Label { text: qsTr("volume") }
```
special: Use title case. This would look good if it was indented slightly on the left, so it's obvious that it is part of "Play audio notifications"

src/utils/AudioPlayer.cpp
```diff
@@ -0,0 +1,83 @@
+#include "AudioPlayer.h"
+#include <QDebug>
+
+//#define USE_QSOUNDEFFECT
+#define USE_MEDIAPLAYER
```
special: On OS X, QMediaPlayer fails (at least while other audio is active?). QSoundEffect seems to work better. I haven't tested against other platforms yet.

I think you mentioned having difficulties with both of them. Can you explain the issues?

special:
Missing the standard license headers in the new .cpp and .h files.

sounds/LICENSE
```diff
@@ -0,0 +1,7 @@
+message.wav, online.wav
+Copyright 2015, qso (qso@users.sf.net)
```
special: Did you create these sounds? Where did they come from originally?

special:
message.wav is pretty decent, but I don't like online.wav very much. It's subtle and difficult to hear/understand. Are there other choices we could look at?

src/utils/AudioPlayer.cpp
qsodev: I have rare problems that ui/application freezes when playing audio using Arch Linux with PulseAudio. 
With QMediaPlayer these hangs occured more often. Using QSoundEffect caused sometimes hangs when in 2 applications the contacts got online simultaniously ( known to each other ).

This issue needs a wider range of testing, preferable also on Windows.

sounds/LICENSE
qsodev: Both sounds are created by myself. 
message.wav is just a short part of the delay feedback of a few drum hits through a filter, all done by myself.
online.wav are recorded leaves of a plant next to me.

special:
This seems to be working pretty well now. Some merge notes (mostly things for me to do):

 * BUILDING needs to mention the new dependency on qtmultimedia and update the package lists
 * I'd like to work on getting sounds that fit better, especially for online.wav. We should file a ticket after merging to think about those.
 * This will likely require a bit of work on the linux static build, but it should be manageable (without requiring things like gstreamer).
 * I squashed and made some very very minor edits in my 188-audio branch
 * Translations need updating

It was pointed out that this could be done with a lot less C++ by using [QML SoundEffect](http://doc.qt.io/qt-5/qml-qtmultimedia-soundeffect.html). That would be a better approach, in my opinion. I'm still okay with merging this as-is (unless you're interested in trying to implement it that way now?), but we should file a ticket to do that eventually.

PR #192, opened 2015-05-27
Allow public keys up to 150 bytes (DER)

### LINKED ISSUE #186: Outbound connections fail when using a vanity address
**NB**: Triggering this issue requires manually replacing Ricochet's hidden service key. Users sticking to the GUI won't see this. In fact, most users shouldn't mess with the internal config files at all.

Ricochet 1.1.0 asserts that its hidden service's DER-encoded public key is 140 bytes long. This seems to be true of keys generated by Tor, but vanity keys generated by software like [Scallion](https://github.com/lachesis/scallion) have large public exponents that require at least an extra byte.

To reproduce, try borrowing the PrivateKey from the example output on the Scallion page (obviously do not use this key for actual conversations) or generate your own and drop it into `config/data-0/private_key`.

When using a release build, this will silently prevent me from connecting to other clients. Note that other clients can still connect to me, so if I'm the only one using a vanity address, the only noticeable effect is that no one will ever receive my friend requests :'(

Using a debug build throws the following and dies when connecting to someone else (ricochet:tgicorjvtqbhcrjt) after starting the client:

> Attempting reconnection of socket to "tgicorjvtqbhcrjt.onion" 9878
> Granting 0 authentication as "tgicorjvtqbhcrjt.onion" to connection
> BUG: Unexpected size for encoded public key
> ASSERT: "something broke!" in file [src/protocol/AuthHiddenServiceChannel.cpp, line 198](https://github.com/ricochet-im/ricochet/blob/master/src/protocol/AuthHiddenServiceChannel.cpp#L198)
> Abort trap: 6

Adding `qDebug() << publicKey.toBase64() << publicKey.size();` before line 198 will dump a copy of the encoded public key and show that it's 141 bytes long.

Temporary fix: Replace the comparison on line 197 with `size() > 150`. I don't know if this is a sane value or if some other validation should be done, but I can now be successfully reached at `ricochet:gabedwrds3khg2uw`.

gabedwrds:
~~Increases the limit on our local key to match validation for partner's key in an incoming connection~~

Sets the limit for both directions to 150 bytes to allow slightly unusual but still valid keys

Fixes #186

### DIFF
--- a/src/protocol/AuthHiddenServiceChannel.cpp
+++ b/src/protocol/AuthHiddenServiceChannel.cpp
@@ -194,7 +194,7 @@ void AuthHiddenServiceChannel::sendAuthMessage()
     }
 
     QByteArray publicKey = d->privateKey.encodedPublicKey(CryptoKey::DER);
-    if (publicKey.size() != 140) {
+    if (publicKey.size() > 150) {
         BUG() << "Unexpected size for encoded public key";
         closeChannel();
         return;
@@ -282,7 +282,7 @@ void AuthHiddenServiceChannel::handleProof(const Data::AuthHiddenService::Proof 
     CryptoKey publicKey;
     if (signature.size() != 128) {
         qWarning() << "Received invalid signature (size" << signature.size() << ") on" << type();
-    } else if (publicKeyData.size() > 180) {
+    } else if (publicKeyData.size() > 150) {
         qWarning() << "Received invalid public key (size" << publicKeyData.size() << ") on" << type();
     } else if (!publicKey.loadFromData(publicKeyData, CryptoKey::PublicKey, CryptoKey::DER)) {
         qWarning() << "Unable to parse public key from" << type();


rburchell:
Is there any actual proscribed limit to this, or should this be increased to something more akin to a sanity check?

gabedwrds:
It's basically a question of how much encoded data there can be in a public key, specifically the public exponent since everything else is fixed. Up to 4 bytes of exponent is common enough that any software should handle it (Ricochet currently expects exactly 3). 8 is crazy large but an irrelevant NIST standard suggests it's at least possible, and anything beyond that is not worth considering.

Some back-of-napkin calculations later, I feel that 150 bytes covers anything we should ever see, and anything over 180 probably violates intergalactic law. I don't know why @special picked that value, but it turns out to be a pretty decent sanity check.

Also, if we wanted to use a different value, it would literally double the LOCs that need to be changed.

src/protocol/AuthHiddenServiceChannel.cpp
```diff
@@ -194,7 +194,7 @@ void AuthHiddenServiceChannel::sendAuthMessage()
     }
 
     QByteArray publicKey = d->privateKey.encodedPublicKey(CryptoKey::DER);
-    if (publicKey.size() != 140) {
```
1407brown: Delete

src/protocol/AuthHiddenServiceChannel.cpp
```diff
@@ -194,7 +194,7 @@ void AuthHiddenServiceChannel::sendAuthMessage()
     }
 
     QByteArray publicKey = d->privateKey.encodedPublicKey(CryptoKey::DER);
```
1407brown: Delete all file

src/protocol/AuthHiddenServiceChannel.cpp
```diff
@@ -194,7 +194,7 @@ void AuthHiddenServiceChannel::sendAuthMessage()
...
-    if (publicKey.size() != 140) {
+    if (publicKey.size() > 180) {
         BUG() << "Unexpected size for encoded public key";
         closeChannel();
         return;
```
1407brown: DLete publicKey

special:
@gabedwrds's logic is sound (and better-researched than was actually necessary). I picked 180 using the metric of "that's n more than any value I've seen so far". The intent of the size check is to avoid ASN.1 parsing anything that is definitely not going to be a well-formed key, in case it's maliciously formed.

This patch looks good. It's matching the outbound sanity check to the inbound one (which is already 180). I accept your reasoning for 150 being a reasonable maximum, so I also approve of a patch which changes both to 150. I'll merge this in either state in the next day or two. Thanks!

gabedwrds:
Yep. Between those comments, I felt a need to research a little deeper after a brief fear that 180 actually might not be enough. After working it out, though, I'm more confident that 150 is plenty. I'll likely amend this to use that value for both directions.

src/protocol/AuthHiddenServiceChannel.cpp
1407brown: Delete

src/protocol/AuthHiddenServiceChannel.cpp
```diff
@@ -282,7 +282,7 @@ void AuthHiddenServiceChannel::handleProof(const Data::AuthHiddenService::Proof
     CryptoKey publicKey;
     if (signature.size() != 128) {
```
1407brown: CLose channel

PR #195, opened 2015-06-09
ui: Copy to the X11 selection clipboard as well

### LINKED ISSUE #194: X11 middle-click paste doesn't see text copied from Ricochet
Roger has discovered that when using X11, with the (default) middle-click-to-paste behavior enabled, the following has a bug:

1. Press "Copy" on next to a contact address
2. Paste by middle-clicking in another field or window
3. Paste by pressing ctrl+v

We expect steps 2 and 3 to paste the address, but step 2 will paste something else instead. This is because middle-click uses the X11 selection buffer, and the contact field doesn't copy into that one. "Copy link" exhibits the same problem.

special:
Fixes #194

### DIFF
--- a/src/ui/LinkedText.cpp
+++ b/src/ui/LinkedText.cpp
@@ -76,6 +76,10 @@ QString LinkedText::parsed(const QString &input)
 
 void LinkedText::copyToClipboard(const QString &text)
 {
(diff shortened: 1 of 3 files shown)


rburchell:
:+1:

PR #196, opened 2015-06-09
Language settings support

qsodev:
added support to change the language


src/main.cpp
```diff
@@ -310,6 +312,16 @@ static void initTranslation()
         qDebug() << "Forcing locale" << locale << "from environment" << locale.uiLanguages();
     }
 
+    SettingsObject settings;
+    QString settingsLanguage( settings.read("ui.language").toString() );
+
+    if (!settingsLanguage.isEmpty())
+    {
```
special: Brace on the same line as if

src/main.cpp
```diff
@@ -310,6 +312,16 @@ static void initTranslation()
...
+        locale = settingsLanguage;
+    } else {
+        settings.write(QStringLiteral("ui.language"), QStringLiteral(""));
```
special: It's not necessary to write the setting when empty - it can be left undefined.

src/ui/LanguagesModel.cpp
```diff
@@ -0,0 +1,77 @@
+/* Ricochet - https://ricochet.im/
+ * Copyright (C) 2014, John Brooks <john.brooks@dereferenced.net>
```
special: If you like, you can put your own name here. You retain copyright on the code you submit.

src/ui/LanguagesModel.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <QDir>
+#include <QLocale>
+#include <QVariant>
+
+LanguagesModel::LanguagesModel(QObject* parent)
+    : QAbstractListModel(parent)
+{
+    // create the list of languages based on present translation files in ":/lang" folder
+    QDir languagesFolder(QStringLiteral(":/lang"));
+    languages.append(LanguageEntry(tr("System"), QStringLiteral("")));
```
special: How about "System default"?

src/ui/LanguagesModel.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+int LanguagesModel::rowCount(const QModelIndex &) const
+{
+    return languages.length();
+}
+
+QVariant LanguagesModel::data(const QModelIndex &index, int role) const
+{
```
special: Check that `index.row() >= 0 && index.row() < languages.length()`. It shouldn't ever happen, but it's better to not crash if it does.

src/ui/MainWindow.cpp
```diff
@@ -97,6 +98,7 @@ bool MainWindow::showUI()
     qml->rootContext()->setContextProperty(QLatin1String("torControl"), torControl);
     qml->rootContext()->setContextProperty(QLatin1String("torInstance"), Tor::TorManager::instance());
     qml->rootContext()->setContextProperty(QLatin1String("uiMain"), this);
+    qml->rootContext()->setContextProperty(QLatin1String("languagesModel"), new LanguagesModel(this));
```
special: You should use `qmlRegisterType` instead of context properties, like ContactsModel does. Then, in LanguagePreferences.qml, you can add `import im.ricochet 1.0` and create a `LanguagesModel` as-needed.

special:
After changing languages, there needs to be some indication that you need to restart Ricochet for it to take effect. Here are a few ideas for how we could do that:

 * Use a popup similar to Bubble in [ContactIDField.qml](https://github.com/ricochet-im/ricochet/blob/master/src/ui/qml/ContactIDField.qml)
 * Display some text on the language page along the lines of "Restart Ricochet to apply changes"
 * Have a "Save" or "Apply" button, and show a dialog when pressed
 * Any others?

src/main.cpp
```diff
@@ -71,11 +70,14 @@ int main(int argc, char *argv[])
     QString error;
     QLockFile *lock = 0;
     if (!initSettings(settings.data(), &lock, error)) {
+        initTranslation();
```
rburchell: is there a reason to not just call initTranslation unconditionally before initSettings, instead of two seperate calls?

src/main.cpp
```diff
@@ -310,6 +312,16 @@ static void initTranslation()
         qDebug() << "Forcing locale" << locale << "from environment" << locale.uiLanguages();
     }
 
+    SettingsObject settings;
+    QString settingsLanguage( settings.read("ui.language").toString() );
```
rburchell: style nitpick: drop the spaces on the sides of the parentheses

src/main.cpp
```diff
@@ -310,6 +312,16 @@ static void initTranslation()
...
+
+    if (!settingsLanguage.isEmpty()) {
+        locale = settingsLanguage;
+    } else {
+        //write an empty string to get "System default" language selected automatically in preferences
+        settings.write(QStringLiteral("ui.language"), QStringLiteral(""));
```
rburchell: use QString() instead of QStringLiteral("")

(do test this, but I think it should work)

QStringLiteral will allocate space in the compiled object for a string, which is really wasteful when there isn't actually any data to put in it :)

src/ui/LanguagesModel.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <QVariant>
+
+LanguagesModel::LanguagesModel(QObject* parent)
+    : QAbstractListModel(parent)
+{
+    // create the list of languages based on present translation files in ":/lang" folder
+    QDir languagesFolder(QStringLiteral(":/lang"));
+    languages.append(LanguageEntry(tr("System default"), QStringLiteral("")));
+    foreach( QString translationFile, languagesFolder.entryList())
+    {
```
rburchell: style nitpicks:
* move the brace to cuddle the foreach
    `foreach (...) {`
* move the space from before the type (QString) to after the foreach

Also, you should const & the foreach arg, otherwise you may cause string detaches unnecessarily (wasteful)

src/ui/LanguagesModel.cpp
```diff
@@ -0,0 +1,80 @@
...
+        QString localeID = translationFile.replace(QStringLiteral("ricochet_"), QStringLiteral("")).replace(QStringLiteral(".qm"),QStringLiteral(""));
```
rburchell: Generally speaking, if the API has a QLatin1String overload, you should prefer that to using QStringLiteral (this doesn't copy the string data, and doesn't convert to utf16 unnecessarily).

QString::replace does have those overloads, so:

translationFile.replace(QLatin1String("ricochet_"), ... etc

and same comment about QStringLiteral("") => QString()

src/ui/qml/LanguagePreferences.qml
```diff
@@ -0,0 +1,57 @@
...
+import QtQuick.Layouts 1.0
+import im.ricochet 1.0
+
+ColumnLayout {
+    anchors {
+        fill: parent
+        margins: 8
+    }
+
+    QtObject {
```
rburchell: what's the reason for this, as opposed to adding the previousLanguage property on the ColumnLayout? (avoids an extra set of allocations)

rburchell:
codewise, looks pretty good aside from some trivial details. :)

src/main.cpp
special: Yes; settings needs to be initialized to get the correct translation language.

src/ui/LanguagesModel.cpp
```diff
@@ -0,0 +1,80 @@
...
+        QString nativeName = QLocale(localeID).nativeLanguageName();
```
special: The QLocale data doesn't capitalize many language names, but in this context I think it makes sense to do so. I think it looks better with:

```
nativeName[0] = nativeName[0].toUpper();
```

src/ui/qml/LanguagePreferences.qml
```diff
@@ -0,0 +1,57 @@
...
+                        uiSettings.write("language", localeID)
+                    }
+                }
+            }
+        }
+    }
+
+    Label {
+        id: restartNotification
+        text: qsTr("Restart Ricochet to apply changes!")
```
special: I still think this would feel nicer as a bubble appearing on the selected button. It's awkward here. Centering the text and adding some spacing above would help, but it wouldn't be perfect.

src/ui/qml/LanguagePreferences.qml
qsodev: It is not necessary but it should be a private property. It is not intended to be modified from outside.

src/ui/qml/LanguagePreferences.qml
qsodev: I could not get the Bubble aligned properly. Directly setting x and y properties helps but even then the Bubble is only aligned correctly after changing the languange a second time.
Help appreciated :)

Issue #13, opened 2011-04-16
Add action support to chat


rburchell:
The ability to do a /me (like in other protocols such as jabber and IRC) would be great, as it is often missed.

PR #197, opened 2015-06-10
Add basic action command

arlolra:
Rudimentary implementation of #13 

Would you prefer to add an isActionRole to the model so the message can be styled a little differently?

### DIFF
--- a/src/core/ConversationModel.cpp
+++ b/src/core/ConversationModel.cpp
@@ -277,7 +277,11 @@ QVariant ConversationModel::data(const QModelIndex &index, int role) const
     const MessageData &message = messages[index.row()];
 
     switch (role) {
-        case Qt::DisplayRole: return message.text;
+        case Qt::DisplayRole:
+          if (message.text.startsWith(QStringLiteral("/me "))) {
+            return QStringLiteral("* %1 %2").arg(m_contact->nickname(), message.text.mid(4, -1));
+          }
+          return message.text;
         case TimestampRole: return message.time;
         case IsOutgoingRole: return message.status != Received;
         case StatusRole: return message.status;


src/core/ConversationModel.cpp
```diff
@@ -277,7 +277,11 @@ QVariant ConversationModel::data(const QModelIndex &index, int role) const
     const MessageData &message = messages[index.row()];
 
     switch (role) {
-        case Qt::DisplayRole: return message.text;
+        case Qt::DisplayRole:
+          if (message.text.startsWith(QStringLiteral("/me "))) {
+            return QStringLiteral("* %1 %2").arg(m_contact->nickname(), message.text.mid(4, -1));
```
s-rah: Some thoughts:

If the message is just "/msg " then the resulting message is "* nick " which is a little odd.
 
Also these look like regular messages, it would be nice to do something like src/ui/LinkedText.cpp to alter the look of these messages so they stand out more (so return the  "/me " string here and morph it on the UI end instead of here - this allows some more flexibility later on if we want to add more actions.

rburchell: I think that customising text here is OK (better, even, in a lot of ways. The C++ string operations are a lot more straightforward from an operational & performance standpoint than having to do them in JS). Changing later, if it's required, is pretty straightforward.

For customizing the appearance, you could add a new role (IsActionMessageRole, maybe) using the startsWith check for the value, and on the UI side, change the text color or whatever fanciness you find appropriate.

src/core/ConversationModel.cpp
```diff
@@ -277,7 +277,11 @@ QVariant ConversationModel::data(const QModelIndex &index, int role) const
     const MessageData &message = messages[index.row()];
 
     switch (role) {
-        case Qt::DisplayRole: return message.text;
+        case Qt::DisplayRole:
+          if (message.text.startsWith(QStringLiteral("/me "))) {
```
rburchell: Use QLatin1String here

(wherever Qt APIs have a QLatin1String overload, prefer using it to QString or QStringLiteral, as the former doesn't do any unicode conversion, and is thus less expensive)

rburchell:
Oh, I see you already arrived at the role conclusion here.

I personally think a role would be handy, but it's only actually useful if you also find a neat way to customize the styling. Maybe @special has an idea on how it could look.

special:
Here's what it looks like in iMessage:

messages image 733433931

That approach could work nicely, although it adds another complex case to the message delegate.

I'm okay with this as-is. It would be neat if somebody finds a fancier way to present it eventually.

Issue #198, opened 2015-06-10
RFC: Making Setup Less Daunting


s-rah:
At the moment, first time users are created with a rather ominous dialogue box, in English presenting 2 options.

![image](https://cloud.githubusercontent.com/assets/106626/8075131/2a6eadd2-0ef1-11e5-9926-f561e3577bb3.png)
        
At the risk of infantilsing certain users I am going to call the following words "scary" - **connection**, **Tor**, **network**, **"connect directly"**, **configure**, **settings**.

Ideally we would not need this configuration prompt at all, but for many users who require initial configuration it is probably best left in

I spent some time this evening playing around with the initial network setup screen that pops up on initial start.

Because this is, 1. the first time people will interact with Ricochet and 2. a fairly important configuration screen in it's own right, I wanted to make it a bit pretty and to make it less technically scary for a non-tech user.

![image](https://cloud.githubusercontent.com/assets/106626/8075098/a0231b9a-0ef0-11e5-8cfb-25b10e31f6bd.png)

Please take a look at the screenshot above which has a rough mockup of what I have been thinking about. I'm still not set on the text - and think the wording at the bottom can probably be made simpler - but I think it captures the essence.

I would be interested in peoples opinions on this train of thought, how easy this is to understand, suggestions on wording etc.

Thanks,

Sarah

special:
Yes. I love this, and thank you so much for thinking about it.

------

For some brief background, I based the text and concept on the Tor Browser (pre-4.5), except that "Configure" gives you a single page with a lot of technical configuration choices instead of a wizard. Tor did a [UX sprint](https://trac.torproject.org/projects/tor/wiki/org/meetings/2015UXsprint) earlier this year, which identified a bunch of [usability](https://trac.torproject.org/projects/tor/ticket/10610) [issues](https://trac.torproject.org/projects/tor/ticket/12164). Tor Browser 4.5 improved the UI a bit, but there's more room for improvement.

If we take a step back, I think there are three categories of users, in this order of frequency:

1. Able to make a direct connection to Tor
2. Censored networks that require bridges
3. Intrusive network configuration, but not explicitly blocking tor (e.g. needs proxy, restricted ports)

I think the current UI is okay (but not good) for groups 1 and 3. I'm not too concerned with optimizing usability for group 3 right now, so the biggest improvement would be making sure that group 1 understands what to do.

[...]

PR #201, opened 2015-06-13
Making the Initial Setup Nicer (#198)

s-rah:
This commit makes some updates to the initial setup UI with the aim of
making it a nicer experience for new users.

image

The initial setup is the first thing a new user of Ricochet sees when
they start up the software - it should be a place where we a) reassure
them that the software is working b) help them perform intiial setup
which we can't figure out and c) help the user form a connection to the
software so we are better able to present intuitive messages to the
user when we do need their input e.g. if the software discovers
something which needs the users attention like their connection to the
Tor  network is blocked.

Presenting complex topics, like proxies and bridges, to a first time
user can be daunting and so the idea behind this initial setup is to
dress up  the initial dialogue and present the user with a sane
default "Press Connect".

There are still issues with this initial setup - in particular when there
is additional configuration needed - we need to refine ways to work with
the user and, where we can, offer suggestions in a safe and secure
manner.

Image copyright information: The images in this commit were created by me
and are distributed under CC BY-SA 4.0 (this is conveyed in the image files)
with the exception of the "police officer icon at the center of the censored.png" 
which is from the Symbola font package http://users.teilar.gr/~g1951d/ which
carries the following licensing terms: "Fonts and documents in this site are not 
pieces of property or merchandise items; they carry no trademark, copyright, 
license or other market tags; they are free for any use."


rburchell:
I'd be interested in how it looks on other platforms too, but that generally looks pretty good. :)

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,95 @@ ApplicationWindow {
...
-                text: qsTr("Connect")
-                isDefault: true
-                onClicked: {
-                    // Reset to defaults and proceed to bootstrap page
-                    configPage.reset()
-                    configPage.save()
+                Image {
+                    anchors.horizontalCenter: parent.horizontalCenter
+                    // Ideally we would use the svg here, but I cannot work out a way to get it to
+                    // display as nice as the png does.
```
rburchell: what's the problem here?

(SVGs are kind of a good thing, since then you get retina support, say, for free)

special: One useful hint is to use `sourceSize { width: n; height: n }` at the desired size for SVGs. That should make it render at the right resolution.

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,95 @@ ApplicationWindow {
...
-                horizontalAlignment: Qt.AlignHCenter
-            }
+            Row {
+
+                Column {
+                    spacing:5
+
+                    Image {
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        source: "../icons/connection_ok.svg"
```
special: sourceSize may be useful on these as well

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,95 @@ ApplicationWindow {
...
+                        isDefault: true
+                        onClicked: {
+                            // Reset to defaults and proceed to bootstrap page
+                            configPage.reset()
+                            configPage.save()
+                        }
+                    }
+
+                    Label {
+                        width: 200
```
special: Avoid hardcoding the width; it leads to subtle layout issues and scalability problems. In this case, you could either:

1. Use a RowLayout instead of Row, put `Layout.fillWidth: true` on the Columns
2. Set the Column's width to `(parent.width-1)/2`

In either case, you'd use `width: parent.width` for these items.

special:
Here's OS X/Retina:

ricochet 2x

The text wrapping makes it very crowded! I think this needs a lot more padding everywhere. I think even your screenshot could use more blank space. Don't hesitate to make the window larger and add lots of whitespace if it comes out looking cleaner.

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,95 @@ ApplicationWindow {
...
+                        text: qsTr("This computer's Internet connection is free of obstacles.")
```
special: I've never really liked this phrase. What do you think of "connection is unfiltered"? "connection is free" sounds too much like it's talking about money, I suppose. Hmm.

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,95 @@ ApplicationWindow {
...
+
+                    Button {
+                        anchors.horizontalCenter: parent.horizontalCenter
+                        text: qsTr("Configure")
+                        onClicked: window.openConfig()
+                    }
+
+                    Label {
+                        width: 200
+                        text: qsTr("This computer's Internet connection is censored, filtered, or proxied.")
```
special: maybe "or uses a proxy" would be better than "or proxied"

src/ui/qml/NetworkSetupWizard.qml
s-rah: I'd like to avoid complex terms here especially with negative prefixes. How about "connection is <b>not</b> censored, filtered and does not use a proxy?" to mimic the right hand side. Although I think a positive phrase would be better....

Although there aren't many positive options...clear?  "connection is clear of obstacles".

src/ui/qml/NetworkSetupWizard.qml
s-rah: Thanks I will give that a try.

src/ui/qml/NetworkSetupWizard.qml
special: Hmm, yes, positive is better. "not censored ..." is too similar to the other side.

"This computer's Internet connection is clear"
"This computer's Internet connection is clear and open"
"This computer's Internet connection is not blocked"
"This computer can access the Internet without restrictions"
"This computer can freely access the Internet"

s-rah: > "This computer can freely access the Internet" 

Is probably my favourite there. Close second:"This computer can access the Internet without restrictions"

I'll have a play around and a think.

s-rah:
Updated diff taking into account, comments, new screenshot in commit messages reflecting what it now looks like on Linux (Ubuntu) - would be interested to see how this looks elsewhere. Much more white space now (Maybe too much...I'm not sure...I've gone back and forth).

Also removed hardcoding width and switched to sourceSize which fixed the logo display issue.

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,110 @@ ApplicationWindow {
...
-                text: qsTr("Connect")
-                isDefault: true
-                onClicked: {
-                    // Reset to defaults and proceed to bootstrap page
-                    configPage.reset()
-                    configPage.save()
+                Image {
+                    anchors.horizontalCenter: parent.horizontalCenter
+                    source: "../icons/ricochet.svg"
+                    sourceSize.width: parent.width / 10
```
rburchell: It's not a good idea to use bindings for sourceSize. Every time it changes, the image will be reloaded (so every time parent.width changes..) which can cause a bad UX on resize (bad performance)

src/ui/qml/NetworkSetupWizard.qml
```diff
@@ -93,43 +93,110 @@ ApplicationWindow {
...
+                        wrapMode: Text.Wrap
+                        horizontalAlignment: Qt.AlignHCenter
+                    }
+                }
 
-            Button {
-                anchors.horizontalCenter: parent.horizontalCenter
-                text: qsTr("Configure")
-                onClicked: window.openConfig()
+                Column {
```
rburchell: So this is just supposed to be a spacer?

    Item {
        width: parent.width / 5
        height: 1
    }

should perform the job OK (you just need to assign both a width and height to force it to take space).

rburchell:
Screenshot looks nice. I haven't tried it myself :)

PR #202, opened 2015-06-16
Add qmake configuration option for (more) hardened builds.

isislovecruft:
GCC-4.x builds compile and (seem to) run well.  I'm slightly concerned that the `-fvtable-verify=std` option is not being applied correctly, but I've not yet dug into finding a suitable methodology for determining whether a class's virtual method table is indeed being protected.  FWIW, I've surprisingly had no crashes yet with my hardened Ricochet build (using GCC-4.9.2)… but then again, I've only got one buddy and only had one conversation.

I haven't yet done sufficient testing of GCC-5.x builds and their additional UBSan flags.

### DIFF
--- a/ricochet.pro
+++ b/ricochet.pro
@@ -40,6 +40,46 @@ CONFIG += c++11
 
 VERSION = 1.1.0
 
+CONFIG(hardening):unix:!macx {
+    message("Using additional compiler flags for hardening purposes.")
+
+    SANITIZER_FLAGS = -fsanitize=address -fsanitize=undefined -fsanitize=integer-divide-by-zero -fvtable-verify=std
+
+    system(which g++-5) | system(g++ --version 2>/dev/null | grep -e "5.[0-9]") {
+        message( "g++ version 5.x found" )
+        QMAKE_CC = $(shell which g++-5)
+        QMAKE_CXX = $(shell which g++-5)
+        SANITIZER_FLAGS += -fsanitize=bounds -fsanitize=vptr -fsanitize=object-size -fsanitize=alignment -fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow
+    }
+
+    # If hardening-wrapper is installed, then use the configured flags
+    exists(/usr/share/man/man1/hardening-wrapper.1.gz) {
+        message("Gathering compiler hardening flags from hardening-wrapper installation.")
+        L_CPPFLAGS = $(shell dpkg-buildflags --get CPPFLAGS)
+        L_CFLAGS   = $(shell dpkg-buildflags --get CFLAGS)
+        L_CXXFLAGS = $(shell dpkg-buildflags --get CXXFLAGS)
+        L_LDFLAGS  = $(shell dpkg-buildflags --get LDFLAGS)
+    }
+    # !exists(/usr/share/man/man1/hardening-wrapper.1.gz) {
+    else {
+        L_CPPFLAGS = -fstackprotector-strong -fPIC -pie -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security -dumpmachine -Wall -Wsecurity
+        L_CFLAGS   = $${L_CPPFLAGS} -g -O2
+        L_CXXFLAGS = $${L_CFLAGS}
+        L_LDFLAGS  = $${L_CPPFLAGS} -Wl,-z,relro
+    }
+    L_LDFLAGS *= -Wl,-z,relro,-z,now
+
+    QMAKE_CPPFLAGS *= $${L_CPPFLAGS} $${SANITIZER_FLAGS}
+    QMAKE_CFLAGS   *= $${L_CFLAGS} $${SANITIZER_FLAGS}
+    QMAKE_CXXFLAGS *= $${L_CXXFLAGS} $${SANITIZER_FLAGS}
+    QMAKE_LDFLAGS  *= $${L_LDFLAGS} $${SANITIZER_FLAGS}
+
+    message(QMAKE_CPPFLAGS $$[QMAKE_CPPFLAGS])
+    message(QMAKE_CFLAGS $$[QMAKE_CFLAGS])
+    message(QMAKE_CXXFLAGS $$[QMAKE_CXXFLAGS])
+    message(QMAKE_LDFLAGS $$[QMAKE_LDFLAGS])
+}
+
 # Pass DEFINES+=RICOCHET_NO_PORTABLE for a system-wide installation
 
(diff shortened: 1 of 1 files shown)


ricochet.pro
```diff
@@ -40,6 +40,46 @@ CONFIG += c++11
 
 VERSION = 1.1.0
 
+CONFIG(hardening):unix:!macx {
```
rburchell: nitpick: maybe "hardened" is a better name than "hardening"?

rburchell: also, you could make this specific to the compiler, rather than the platform, using something like this to replace the "unix:!macx" construct:

*-g++* (not *-g++-*, though that would work on e.g. Linux, because you then won't match e.g. netbsd-g++)

(please test this though, that's just off the top of my head)

ricochet.pro
```diff
@@ -40,6 +40,46 @@ CONFIG += c++11
...
+    message("Using additional compiler flags for hardening purposes.")
+
+    SANITIZER_FLAGS = -fsanitize=address -fsanitize=undefined -fsanitize=integer-divide-by-zero -fvtable-verify=std
+
+    system(which g++-5) | system(g++ --version 2>/dev/null | grep -e "5.[0-9]") {
```
rburchell: this makes me really uneasy, and I don't think you should do it. it would break cross compilation totally, as well as any other use of a non-system-installed toolchain (which in qmake's case would be using a different mkspec)..

I'd recommend not touching QMAKE_CC/CXX, and looking into version detection of the g++ version by using system() on QMAKE_CC or something to add to the sanitize flags (which is not really portable, but slightly less broken at least)

ricochet.pro
```diff
@@ -40,6 +40,46 @@ CONFIG += c++11
...
+    }
+
+    # If hardening-wrapper is installed, then use the configured flags
+    exists(/usr/share/man/man1/hardening-wrapper.1.gz) {
+        message("Gathering compiler hardening flags from hardening-wrapper installation.")
+        L_CPPFLAGS = $(shell dpkg-buildflags --get CPPFLAGS)
+        L_CFLAGS   = $(shell dpkg-buildflags --get CFLAGS)
+        L_CXXFLAGS = $(shell dpkg-buildflags --get CXXFLAGS)
+        L_LDFLAGS  = $(shell dpkg-buildflags --get LDFLAGS)
+    }
```
rburchell: for style, i'd drop the comment and just cuddle the brace:

`} else {`

ricochet.pro
```diff
@@ -40,6 +40,46 @@ CONFIG += c++11
...
+
+    system(which g++-5) | system(g++ --version 2>/dev/null | grep -e "5.[0-9]") {
+        message( "g++ version 5.x found" )
+        QMAKE_CC = $(shell which g++-5)
+        QMAKE_CXX = $(shell which g++-5)
+        SANITIZER_FLAGS += -fsanitize=bounds -fsanitize=vptr -fsanitize=object-size -fsanitize=alignment -fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow
+    }
+
+    # If hardening-wrapper is installed, then use the configured flags
+    exists(/usr/share/man/man1/hardening-wrapper.1.gz) {
```
rburchell: this is probably also not good from a xcompile perspective, but YOLO!

ricochet.pro
```diff
@@ -40,6 +40,46 @@ CONFIG += c++11
...
+        L_LDFLAGS  = $${L_CPPFLAGS} -Wl,-z,relro
+    }
+    L_LDFLAGS *= -Wl,-z,relro,-z,now
+
+    QMAKE_CPPFLAGS *= $${L_CPPFLAGS} $${SANITIZER_FLAGS}
+    QMAKE_CFLAGS   *= $${L_CFLAGS} $${SANITIZER_FLAGS}
+    QMAKE_CXXFLAGS *= $${L_CXXFLAGS} $${SANITIZER_FLAGS}
+    QMAKE_LDFLAGS  *= $${L_LDFLAGS} $${SANITIZER_FLAGS}
+
+    message(QMAKE_CPPFLAGS $$[QMAKE_CPPFLAGS])
```
rburchell: might be an idea to stick these behind a hardening_debug option, since the full set of options will be visible on compiles anyway

rburchell:
interesting work, thanks!

I've made some comments on the approach, I hope you don't mind :-)

ricochet.pro
special: Still needs `unix:` to rule out win32-g++ (a mingw spec).

ricochet.pro
```diff
@@ -40,6 +40,46 @@ CONFIG += c++11
...
+    exists(/usr/share/man/man1/hardening-wrapper.1.gz) {
+        message("Gathering compiler hardening flags from hardening-wrapper installation.")
+        L_CPPFLAGS = $(shell dpkg-buildflags --get CPPFLAGS)
+        L_CFLAGS   = $(shell dpkg-buildflags --get CFLAGS)
+        L_CXXFLAGS = $(shell dpkg-buildflags --get CXXFLAGS)
+        L_LDFLAGS  = $(shell dpkg-buildflags --get LDFLAGS)
+    }
+    # !exists(/usr/share/man/man1/hardening-wrapper.1.gz) {
+    else {
+        L_CPPFLAGS = -fstackprotector-strong -fPIC -pie -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security -dumpmachine -Wall -Wsecurity
```
special: We could enable most of this by default. I think `-fstackprotector-strong` is the only exception (because of the version requirement), unless some of the warnings are also too new.

special:
I agree with @rburchell's comments, but otherwise, I'm all for merging this behind an opt-in flag - and going further, putting anything safe and well-supported on by default. Shame that qmake doesn't offer a nicer way to detect compiler versions.

ioerror:
I'd encourage enabling every hardening option possible by default. It might make sense to allow opt-out options that ensure that users understand they're voiding any and all security expectations.

I'd also encourage the use of AddressSanitizer with clang.

ioerror:
When trying the first patch by @isislovecruft - I can't actually get the asan headers and references to lead me to a compile. Furthermore, `hardening-check` doesn't report that those flags result in a build that looks very hardenend. :-(

ioerror:
For those playing along at home, you'll need to apply the patch and then run qmake like so:
`qmake DEFINES+=RICOCHET_NO_PORTABLE CONFIG+=hardening`
Then the usual `make` and `sudo make install`

special:
These changes were merged as part of #229. Thanks!

PR #203, opened 2015-06-17
Improve window alert behavior on OS X

### LINKED ISSUE #145: Dock alerts are annoying on OS X
When you receive a message on OS X, the window alert causes the dock icon to bounce. It will never stop bouncing until you focus the window. That can be pretty annoying, and it's not the correct behavior; most applications seem to bounce once or twice, and then stop.

special:
Fixes #145

### DIFF
--- a/src/core/ContactsManager.cpp
+++ b/src/core/ContactsManager.cpp
@@ -38,6 +38,10 @@
 #include <QStringList>
 #include <QDebug>
 
+#ifdef Q_OS_MAC
+#include <QtMac>
+#endif
+
 ContactsManager *contactsManager = 0;
 
 ContactsManager::ContactsManager(UserIdentity *id)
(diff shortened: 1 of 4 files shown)


src/core/ContactsManager.cpp
```diff
@@ -186,3 +186,13 @@ void ContactsManager::onUnreadCountChanged()
     emit unreadCountChanged(user, model->unreadCount());
 }
```
rburchell: Why do you have a signal here, listened to in QML, that calls back to get the count here, then goes back to QML, to call into C++ in MainWindow just to use MacExtras, instead of adding something like this after the emit (or even connected C++ side if you want to be super safe):

```
    int re = 0;
    foreach (ContactUser *u, pContacts) {
        if (u->conversation())
            re += u->conversation()->unreadCount();
    }
#ifdef Q_OS_MAC
    QtMac::setBadgeLabelText(QString::number(re));
#endif

rburchell: I mean, I'm all for "encapsulation" as a reason, but in my opinion this would be a lot more straightforward to follow.

rburchell:
Looks good, though I subjectively think it could be better :)

rburchell:
:+1:

PR #205, opened 2015-06-20
Assorted minor changes from developing file transfers

special:
These commits are all independent, and mostly have no effect on existing code. They are all extracted from the upcoming file transfers branch.


rburchell:
Few comments. Looks good on the whole.

rburchell:
:+1:

PR #229, opened 2015-07-25
Use compiler tests to enable hardening flags when available

special:
Based on @isislovecruft's #202, I found a way to get qmake to test for compiler support, so we can maybe have these on by default.

I omitted the debian hardening-wrapper stuff. I am skeptical that this is the right place to apply it.

Someone should think about which options we should try to use, and in what groupings. This also needs testing in many places to make sure we're not breaking something.


hardened.pri
```diff
@@ -0,0 +1,22 @@
+load(configure)
+# Define common variables; these are used by config tests _and_ the actual build
+HARDENED_SANITIZE_FLAGS = -fsanitize=address
+# Also: -fsanitize=undefined -fsanitize=integer-divide-by-zero -fvtable-verify=std
```
special: These are available in GCC 4.9, but not 4.8. Maybe they should be combined with SANITIZE_MORE? Can somebody research this?

hardened.pri
```diff
@@ -0,0 +1,22 @@
...
+HARDENED_SANITIZE_MORE_FLAGS = -fsanitize=bounds -fsanitize=vptr -fsanitize=object-size -fsanitize=alignment -fsanitize=float-divide-by-zero -fsanitize=float-cast-overflow
+HARDENED_STACKPROTECTOR_FLAGS = -fstackprotector-strong
```
special: Oh, I bet this is supposed to be `-fstack-protector-strong`.

hardened.pri
isislovecruft: What about having three categories, vaguely: GCC<=4.8, GCC-4.9, and GCC-5.x?

rburchell:
Looking better. I have no comments so far, I'll leave it to @isislovecruft

special:
I did some brief research into the sanitizer options, and I no longer think we should enable them for release builds. They are more useful as bug-finding tools:

##### AddressSanitizer might have significant runtime impact

For example, check out the 'Limitations' section [in clang docs](http://clang.llvm.org/docs/AddressSanitizer.html#limitations):

> AddressSanitizer uses more real memory than a native run. Exact overhead depends on the allocations sizes. The smaller the allocations you make the bigger the overhead is.
>
> AddressSanitizer uses more stack memory. We have seen up to 3x increase.
>
> On 64-bit platforms AddressSanitizer maps (but not reserves) 16+ Terabytes of virtual address space. This means that tools like ulimit may not work as usually expected.

##### UBSan isn't fatal by default

Default behavior for the undefined behavior sanitizer is to print and continue in most cases. That's toggled with the `-fno-sanitize-recover` flag, but I'm worried about stability if we enabled that on release builds without testing first.

-------

I suggest we enable the sanitizers by default for debug builds, and opt-in for release builds.

Stack protector and relro are always good. Anything else we should add? @ioerror?

ioerror:
I think you should enable by default - the chances that AddressSanitizer is the security or reliability problem is nearly zero when Ricochet is currently crashing for people. Especially if we consider the security of a program written in a non-type safe language. It would be amazing if running with ASAN was the worst problem. That is a problem you want to have and I think is unlikely to happen in our lifetimes. : )

( Also `PIE`, relro, `-fno-sanitize-recover`, binding now, etc)

special:
@ioerror there are two potential problems: 1) that ASAN/UBSAN will cause crashes _for end-users_ in situations that weren't actually dangerous; and 2) that ASAN's overhead and behavior causes problems on some end-user systems, such as excessive memory use. We're not too worried about performance here, fortunately.

I need to have some idea how big/likely those problems are before merging it for release builds by default. Is anybody else shipping software/distributions that use the sanitizers by default?

ioerror:
@special 

If you merge this with a debug flag, I'll run it and give you feedback. I don't think that 1) is likely as Ricochet already crashes in dangerous ways and this will probably stop it or allow us to debug it while keeping them mostly safe from harm. As far as 2) - I don't think it will cause excessive memory *usage* - it tells the kernel what it wants but the kernel (on GNU/Linux) is smarter and doesn't give it what it wants, only what is used. At least that is what I understand - so it shouldn't matter at all.

My understanding is that some people run Chrome with ASAN which is way way more complicated and actually has to be performant.

Issue #226, opened 2015-07-25
Audio notifications sometimes play when they shouldn't


special:
I've heard the "contact online" sound play when I definitely have that option disabled. It might happen just after startup, or in some other condition. I'm not quite sure yet.

Relatedly, check whether this code initializes audio at all when the option is disabled. It's good to avoid that when possible for tasks like #222.

PR #230, opened 2015-07-26
ui: Use QML SoundEffect for audio notifications

special:
This is a simpler implementation, and fixes a bug that sometimes
caused sounds to play on startup despite being disabled.

This also ensures that no audio resources are used unless this
option is enabled, which is useful for sandboxes and other
restrictive environments.


special:
This fixes #226.

rburchell:
:+1: 

I'm a bit unsure about the Loader use. Will errors in the dynamically loaded stuff still show correctly?

special:
Yes, if/when that option is enabled.

It's useful to prevent any audio connection from happening unless that is turned on.

PR #232, opened 2015-07-27
Sandbox contrib directory

ioerror:
This is the start of some useful files for sandboxing on Debian GNU/Linux.


special:
The `contrib/ricochet-seccomp-*.policy` files are all 0 bytes. Is that intentional?

ioerror:
The policy files are set now - amd64 is the tested policy. The others are just basic policy files that mirror it and will probably need to be tweaked.

Issue #22, opened 2014-03-28
Packaging for common Linux distributions


special:
We want Ricochet packaged for:
- [x] Debian
- [x] Ubuntu
- [x] Fedora (and other redhat systems)
- [x] Arch

In addition to writing the packages, each of these needs someone that is willing to volunteer to keep them up to date.

cicku:
https://bugzilla.redhat.com/show_bug.cgi?id=1264700 , I'm the reviewer.

But I still don't trust ricochet, not even tor/tox . :joy:

special:
Ricochet is packaged for debian, ubuntu, fedora, and arch, thanks to package maintainers for each of those platforms.

PR #234, opened 2015-07-27
Added basic RPM spec for CentOS/Fedora/...

pludi:
Change contains a basic SPEC to build & package for RPM based distributions. Can be used to create RPMs for download from the main page, as well as a starting point for distribution-specific SPECs.

Covers another case of #22.


special:
Thanks!

PR #239, opened 2015-07-28
Build files for Debian/*buntu

pludi:
This commit contains the bare minimum to build proper DEBs, bringing #22 yet another step forward (hopefully).


packaging/debian/control
```diff
@@ -0,0 +1,20 @@
+Source: ricochet
+Maintainer: Peter Ludikovsky <peter@ludikovsky.name>
+Section: misc
+Priority: optional
+Standards-Version: 3.9.6
+Build-Depends: debhelper (>=9), libssl-dev (>=1.0.0), pkg-config, gcc (>=4.9) | clang (>=3.5), make (>=4.0), libprotobuf-dev, protobuf-compiler, qt5-qmake (>=5.1.0), qtbase5-dev (>=5.1.0), qtchooser (>=5.1.0), qtbase5-dev (>=5.1.0), qttools5-dev-tools (>=5.1.0), qtdeclarative5-dev (>=5.1.0), qtmultimedia5-dev (>=5.1.0), qml-module-qtquick-controls (>=5.1.0), qml-module-qtquick-dialogs (>=5.1.0), tor (>=0.2.5.12)
```
special: Is it actually necessary to build-depend on `gcc | clang`, `make`, etc?

Also, these:

```
qml-module-qtquick-controls (>=5.1.0), qml-module-qtquick-dialogs (>=5.1.0), tor (>=0.2.5.12)
```

Are not build-dependencies; they should be install dependencies instead.

packaging/debian/ricochet.debhelper.log
```diff
@@ -0,0 +1,2 @@
+override_dh_auto_configure dh_auto_configure
```
special: Seems like a build artifact file? Probably shouldn't be here.

special:
I was only able to get this to build by moving `debian` to the root of the repository and running `dpkg-buildpackage -B`. Is there another way? Should there be a README, or is this obvious to people who are more debian-y than I am?

pludi:
I included those `Build-Depend`s since `lintian` complained about them not being there, and of `build-essential` being a meta-package.
The `override_dh_auto_configure` is an artifact, that's right, from testing `clang`. However, since no override is happening, it shouldn't impact the build process itself.

These files, in their current state, are meant primarily as a starting point for any potential distribution packagers, same as the RPM spec I've provided earlier. Aside from that I'll try to create a small script that'll auto-build in the appropriate way from a cloned repository, but that will come with a separate pull request.

Issue #247, opened 2015-08-13
Tray icon


kolodziej:
I have started developing tray icon for Ricochet. You can check progress at [on **tray-icon** branch of my fork](https://github.com/kolodziej/ricochet/tree/tray-icon). By now it can hide and show main window and show context menu with some actions on right click.

special:
Duplicate of #10

PR #252, opened 2015-08-18
Tray icon

kolodziej:
Features:
1. Left click on tray icon show and hide main window
2. Right click gives context menu
3. When tray icon is activated in settings (general preferences), X button of main window hides it. Otherwise it works as usually - closes the app.
4. Tray icon can show notifications about new messages. I have decided to show only information that user has a new message without showing neither from whom it is nor what is its content.
5. Left click on notification raise main window
6. Right click on notification hides it without showing main window.
7. In general preferences user can set duration in which notification will be shown.
8. There is also option to set notification in 'movie-style'. It was inspired by e-mail notifications which can be often seen in movies. Big, centered on screen. I think it's not just 'fun thing'. I'll use it :)

This PR is related to issue #247


rburchell:
I don't think this is reviewable as-is.

Please take a few minutes to read up on rebasing in git ("git pull --rebase") to place your changes on top of the latest codeline -- http://www.git-scm.com/book/en/v2/Git-Branching-Rebasing

As well as that, you really want to be looking into using separate branches for separate things (the polish language changes likely should have been kept on a separate branch for your other PR, not done as part of this).

Finally, you should look into amending commits when you discover that something needs to be changed in them (git commit --amend, and interactive rebasing -- http://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History) in order to keep the history readable and understandable.

Each pull request should be one logical unit of work (a bugfix, or a feature implementation) and each commit inside a pull request should be an atomic unit of work (like, say, "implement feature X for windows", "implement feature X for mac"). Learning how to split work into commits is a bit of an acquired art rather than something that can be taught, I'd encourage you to look at the commit logs of ricochet and other open source projects (using 'git log' etc) to get an idea of how it should work.

At a brief scan over the diff, there's likely a lot of work to put in before this is in mergable state, but I don't think it's worth trying to get to that point until the pull request/commit history looks sane. Feel free to ask for help, if you need it.

special:
In addition to what @rburchell mentioned, here are a few broad comments on the changes:

Custom popup windows for notifications are problematic. They don't match OS theme or behavior, often steal focus, and in my experience are almost universally annoying. Please do the notifications in a separate pull request - and I think we should look into using OS native ones first.

Having hide-on-close behavior by default is confusing. People will not expect that Ricochet keeps running when they close it, and many won't notice the tray icon.

The changes to how window/dialog close look wrong. At a glance, it looks like no window is ever actually closed and resources are never freed for the dynamically created dialogs. I don't see why any changes would be necessary except for the main window.

What happens with contact requests while hidden?

There's no need to commit the `lupdate` changes along with every code change - it adds a lot of noise. In general, the idea is to run that periodically before releases and sync the results to Transifex.

PR #255, opened 2015-08-21
Update ricochet_es.ts

Vuelos:


special:
Thanks! It looks like there were some updates to Spanish on Transifex that hadn't been merged in yet. Most are the same as your changes, but not all. In particular:

 * "Outdated" is translated as "Caducada"
 * "Select Language" is translated as "Seleccionar idioma"
 * "System Default" is translated as "Predeterminado del sistema"
 * "Language" is translated as "Idioma"

I invited you to join the Spanish team on Transifex - if you disagree with those translations, that would be the best place to raise the issue or change them. If these are regional differences in Spanish, I'll leave it up to you and the other Spanish translators to decide if it's important to have a separate translation for it.

Issue #233, opened 2015-07-27
Fix issues with Qt 5.5


special:
It's new-Qt-release time again! Put the new issues here.

* [x] OS X, "Copy" in context menus does nothing (but cmd+C still works)
* [x] OS X, "Open with Browser" does nothing
* [x] OS X, conversation view doesn't always scroll down when sending messages
* [ ] Lots of debug output like: `AbstractCheckable_QMLTYPE_132 QVariant(Invalid) QRect(0,0 0x0)`

special:
Context menu issues fixed by 516cda4949f7464379ffc11f596192f99099571c.

PR #256, opened 2015-08-25
Use custom scrolling instead of ScrollView

special:
ScrollView is extremely buggy, and seems to regress in every Qt release.
It has a lot of trouble keeping the view scrolled to the bottom, and
with moving at a predictable rate in response to wheel events.

Writing a simple visual scrollbar and wheel handling manually makes for
a much better experience, even if we lose some platform-native scrollbar
behaviors.

---

This could use more testing on Qt <5.4, and on Windows or older Linux systems. It fixes a few Qt 5.5 regressions (#233).


rburchell:
:+1: 

Should insulate from future problems which is nice.

PR #257, opened 2015-08-25
Limit chat message length in UI

special:
This limit is applied by the protocol, so it should be apparent in UI as
well.

Also fixes a QTextCursor warning after sending messages, caused by
setting the text property instead of using remove().

### DIFF
--- a/src/ui/qml/ChatPage.qml
+++ b/src/ui/qml/ChatPage.qml
@@ -144,8 +144,15 @@ FocusScope {
                 }
 
                 function send() {
+                    if (textInput.length > 2000)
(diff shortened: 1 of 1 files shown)


rburchell:
The "setting the text property" thing sounds like a bug to me. I'm OK with change to fix the warning, but can you please report a bug upstream too?

rburchell:
Is there some kind of a visual feedback about the text being cut? This is :+1: though.

special:
There is feedback of a sort, in that your message gets cut off and any further typing doesn't show up. I don't care to add more than that at this point

Issue #74, opened 2014-08-13
Incoming contact request while contact dialog is open freezes UI


special:
Applies to OS X and some Linux systems (gnome 2?):
1. Open "Add Contact" dialog
2. Share address with Bob, leaving the contact dialog open
3. Receive a contact request from Bob

The incoming contact request tries to open a second modal window, and some systems freak out at this. It can result in the entire program being impossible to interact with, or even breaking the window system.

kolodziej:
I have found that removing parent object when creating dialog helps in my case (I use KDE 4.14.2 on Debian Jessie).

PR #259, opened 2015-09-06
Resolve problem #74

kolodziej:
When contact request comes main.qml creates dialog without setting
parent window. It allow to open more contact requests dialogs even
if other dialogs are opened (including AddContactDialog).

I use KDE 4.14.2 and it works for me.


special:
The real fix we need here is to change the UI design to handle contact requests in a less intrusive way - without a popup dialog. This seems to slightly improve it in the interim, though.

Thanks

PR #260, opened 2015-09-06
Ricochet's icon in tray

kolodziej:
Features:
- Icon changes when user has unread messages.
- Left click on tray icon shows and hides main window
- Right click gives context menu with four options: 
  - preferences
  - add contact
  - copy my id [to clipboard]
  - quit
- User can activate option which will change behavior of close ('X') button to make it hides window, not closes application.

---
This PR is related to issue #247


src/ui/MainWindow.cpp
```diff
@@ -67,6 +68,8 @@ MainWindow::MainWindow(QObject *parent)
     Q_ASSERT(!uiMain);
     uiMain = this;
 
+    trayIcon = new TrayIcon(QIcon(QLatin1String(":/icons/ricochet.svg")), QIcon(QLatin1String(":/icons/ricochet_unread.svg")));
```
special: trayIcon is leaked on destruction

src/ui/MainWindow.h
```diff
@@ -33,6 +33,7 @@
 #ifndef MAINWINDOW_H
 #define MAINWINDOW_H
 
+#include "TrayIcon.h"
```
special: This could be a forward declaration instead of including the full header

src/ui/TrayIcon.cpp
```diff
@@ -0,0 +1,103 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "TrayIcon.h"
+
+TrayIcon::TrayIcon(QIcon std_icon, QIcon unread_icon) :
```
special: nitpick: `const &`

src/ui/TrayIcon.cpp
```diff
@@ -0,0 +1,103 @@
...
+        m_std_icon(std_icon),
+        m_unread_icon(unread_icon)
+{
+    connect(this, SIGNAL(activated(QSystemTrayIcon::ActivationReason)), this, SLOT(activated(QSystemTrayIcon::ActivationReason)));
```
special: Use Qt5 style connections, and name the slot something else, since they're both on the same object.

`connect(this, &QSystemTrayIcon::activated, this, &TrayIcon::onActivated);`

src/ui/TrayIcon.cpp
```diff
@@ -0,0 +1,103 @@
...
+    setIcon(m_std_icon);
+
+    m_context_menu = new QMenu();
+    m_context_menu->addAction(tr("Preferences"), this, SLOT(openPreferences()));
+    m_context_menu->addAction(tr("Add contact"), this, SLOT(addContact()));
```
special: Ricochet uses title case in context menus (capitalize Contact).

src/ui/TrayIcon.cpp
```diff
@@ -0,0 +1,103 @@
...
+    m_context_menu->addAction(tr("Copy my ID"), this, SLOT(copyMyId()));
+    m_context_menu->addSeparator();
+    m_context_menu->addAction(tr("Quit"), this, SLOT(quitApplication()));
```
special: Why not connect this directly to `qApp` and `SLOT(quit())`?

src/ui/TrayIcon.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+TrayIcon::TrayIcon(QIcon std_icon, QIcon unread_icon) :
+        m_std_icon(std_icon),
+        m_unread_icon(unread_icon)
+{
+    connect(this, SIGNAL(activated(QSystemTrayIcon::ActivationReason)), this, SLOT(activated(QSystemTrayIcon::ActivationReason)));
+    setIcon(m_std_icon);
+
+    m_context_menu = new QMenu();
+    m_context_menu->addAction(tr("Preferences"), this, SLOT(openPreferences()));
```
special: You can connect directly to signals instead of creating a slot that emits the signal. Just use `SIGNAL(preferences())` here instead.

src/ui/qml/MainWindow.qml
```diff
@@ -19,6 +19,17 @@ ApplicationWindow {
     onMinimumWidthChanged: width = Math.max(width, minimumWidth)
     onMaximumWidthChanged: width = Math.min(width, maximumWidth)
 
+    onClosing: {
+        if (uiSettings.data.hideOnX)
+        {
```
special: nitpick: braces on the same line as if/else

special:
I think the tray icon should only be visible when close-to-tray is enabled. We could also make a separate option, but I like that less. What do you think?

Looking good - I think this can be ready to merge after the above comments.

kolodziej:
Ok. I think everything's done.

src/ui/TrayIcon.h
```diff
@@ -0,0 +1,72 @@
...
+
+class TrayIcon : public QSystemTrayIcon
+{
+    Q_OBJECT
+    Q_PROPERTY(bool unread WRITE setUnread)
+public:
+    TrayIcon(const QIcon& std_icon, const QIcon& unread_icon);
+    ~TrayIcon();
+
+    void setStdIcon(QIcon std_icon);
```
rburchell: these are unused?

src/ui/qml/main.qml
```diff
@@ -95,6 +94,32 @@ QtObject {
             }
         },
 
+        Connections {
+            target: trayIcon
+            onToggleWindow: {
+                if (mainWindow.visibility == Window.Hidden ||
+                    mainWindow.visibility == Window.Minimized)
+                {
```
rburchell: ) {, and

{ else } below

src/ui/TrayIcon.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void TrayIcon::resetIcon()
+{
+    setIcon(m_std_icon);
+}
+
+void TrayIcon::setUnread(bool unread)
+{
+    if (unread) setIcon(m_unread_icon);
```
rburchell: newline after ) and else

kolodziej:
I hope now it's ok :)

rburchell:
thanks :)

qsodev:
looks good :)

kolodziej:
It's about time to **merge** :)

puzanov:
Is it in master?

kolodziej:
Not yet, but it's ready since end of September.

special:
Toggling the window when the tray icon is clicked doesn't behave well. On some platforms, the window is hiding every time you click to open the context menu on the tray. How do most messaging-type applications behave here?

It's less important, but the unread icon is too subtle, also. Maybe recoloring the entire icon to orange would be better. We could merge without that.

We should disable tray icons entirely on OS X. Adding a `#ifndef Q_OS_MAC` in the right place would do. I can do this after merge if needed.

qsodev:
For example Pidgin and Psi open and close the main window on single click. Right click opens the context menu.

kolodziej:
I turned off icon on OS X, but there is an easy to resolve conflict in `src/ui/MainWindow.cpp`. Can I do it on my own or I should left it as it is?

ZimTester:
Single click is the best Windows practice, agree

kolodziej:
The only think that could be done is new 'unread icon'. After merging this branch into master we can make an issue. Maybe we have anybody who is better in graphics than me ;)

qsodev:
what about simply flashing the already existing tray icon when there are unread messages?

kolodziej:
I think it could be done. What do you think, should it be default behaviour or just an option?

ZimTester:
I still think the best way is using a special icon for unread incoming messages, and here is why:
Flashing should be switchable on/off by option, because not everybody likes it. But it's ok to make it on by default. But when it's off we should see that we have unread messages, that's why we need a special icon. (Also if we have it then we should see the same special icon when flashing is on too).
Maybe you can choose something from any free icon packs? Hm...

kolodziej:
We have special icon with dot. It's not very effective and I wanted to change it. I didn't know if it is enough to set new, more different icon or make it flash.

kolodziej:
I've changed unread icon and add blinking when we have unread messages. I think now this feature is ready to merge.

PR #261, opened 2015-09-08
Add .travis-ci.yml for Linux

kolodziej:
.travis.yml contains proper configuration of Travis-CI service
for Linux builds in Debug and Release configurations. README.md
has icon for with current building status. Icon indicates status
for https://github.com/ricochet-im/ricochet (branch master)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,34 @@
+language: cpp
+
+os:
+  - linux
+
+env:
+  - CONFIG=Debug
+  - CONFIG=Release
+
+install:
+  - sudo apt-add-repository -y ppa:ubuntu-toolchain-r/test
+  - sudo apt-add-repository -y ppa:beineri/opt-qt541
+  - sudo apt-get -qq update
+  - sudo apt-get install -qq build-essential libssl-dev pkg-config
+  - sudo apt-get -qq install g++-4.8 libc6-i386 qt54tools qt54svg qt54multimedia qt54declarative qt54quick1 qt54quickcontrols qt54translations qt54imageformats 
+  - git clone https://github.com/google/protobuf.git
+  - cd protobuf
+  - git checkout v2.6.1
+  - ./autogen.sh
+  - ./configure
+  - make
+  - make check
+  - sudo make install
+  - cd ..
+  - sudo ldconfig
+  - export CXX="g++-4.8"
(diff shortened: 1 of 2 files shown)


special:
Hmm..

Automated builds would be a great thing to have. I haven't considered Travis so far, because they don't have many platforms (just one old version of Ubuntu and OS X, I think?). It seems like this would only be useful for catching compilation errors on Linux and that version of Qt. What do you think?

It would be more interesting if we didn't have to pull in a specific Qt from some PPA. Ideally, I'd love automated builds using the [buildscripts](https://github.com/ricochet-im/buildscripts) for all platforms - which isn't hard to set up with Jenkins or Buildbot, but it'd take a few days of wrangling systems, and it's complicated because Windows and OS X can't be cross-compiled.

kolodziej:
You're right. If we had server with KVM on which we could install more virtual systems (including Windows and OS X) we could have great build system which will produce binaries and deploy them in few minutes. Due to problems with Qt version on Travis I had to install Qt and compile Google Protocol Buffers' compiler each time (it takes about 7 minutes when entire process takes 11!). I've also tried to configure AppVeyor which makes Windows compilation but it is way harder task (probably due to the fact that I don't use Windows for programming for many years and don't know build tools other than DevCpp :)).

xor-gate:
Why not use DigitalOcean VPS API it is cheap as crazy because you pay per-minute usage. You can spinup a VM and create build a release, publish the artifact and then destroy it. They support Debian, FreeBSD etcetera.

dguido:
You can probably ditch the PPAs if you switch to the Travis Ubuntu 14.04 infrastructure. Put this at the top of the Travis file:

```
sudo: required
dist: trusty
```

@special: Travis is useful because you can have it run tests after the build and report the output, not just catch compiler errors. Travis integrates with Github too, so it will do a build on every pull request that gets open and give the submitter immediate feedback if they broke anything (ie. tests don't pass).

special:
Sorry about letting this bitrot for so long. I had aspirations for awhile of getting a build infrastructure up that would match target builds/platforms more closely, but it never panned out because, it turns out, that's a lot of work.

I've fixed this up to get it working again, added osx, and submitted it as #513. Let's move over there, thoughts would be very appreciated.

PR #262, opened 2015-09-08
added tray icon support

qsodev:
support for a trayicon:
minimal context menu ( hide/ show and quit )
message icon when message received
support for trayicon text popup messages ( currently not enabled )


special:
Thanks for doing this, but it looks like there was a collision of effort. I think we'll take #260 instead; it's simpler, and I like that it keeps more of the UI logic within QML. Could you take a look at that PR and see if you have any comments?

One notable difference is that you also have tray notification support. That's something we should do, but it needs to be a separate pull request, because it has a few complications: in particular, we have to make sure that the operating system doesn't end up recording anything we don't want to disk. I think it will also need to be behind an option.

src/ui/qml/MainWindow.qml
```diff
@@ -39,16 +39,21 @@ ApplicationWindow {
             if (unreadCount > 0) {
                 if (audioNotifications !== null)
                     audioNotifications.message.play()
+                trayIcon.notifyMessage()
+                if (trayIcon.messagesEnabled) trayIcon.showText(qsTr("Event"), qsTr("New Message from")+" "+user.nickname)
```
special: For future reference: you can't append/prepend text to translated strings, because it might not translate in the same order in all languages. Qt supports placeholders for this, like:

`qsTr("New message from %1").arg(user.nickname)`

PR #264, opened 2015-09-15
Allow access to gtkrc-2.0

Phoul:
Add "owner @{HOME}/.gtkrc-2.0 r," to apparmor profile, giving Ricochet access to GTK theme information.


special:
Thanks!

I think a lot of the pulseaudio rules in this profile aren't actually necessary. Someone should look into that as well.

Issue #241, opened 2015-07-31
A user experiences frequent crashes on OS X


special:
Someone has reported fairly frequent crashing with 1.1.0 on OS X Mavericks. I haven't heard reports of similar crashes from other users yet.

They mention that it sometimes happens when switching between applications with a couple conversation windows open. I have backtraces, but they are so far not helpful.

special:
Haven't heard more reports of this.

PR #266, opened 2015-09-16
Use QSharedPointer for Connection ownership

special:
This makes the ownership of a Connection more clear, and removes a
lot of sketchy use of deleteLater to try to ensure connections are
always cleaned up at the right times.


special:
This may be related to #241.

PR #270, opened 2015-09-23
Certificats change

burdges:
Appears Qt's certificates have changes and qt-project.org no longer works over TLS.  It's possible https://www.qt.io/download-open-source/ is preferable.


burdges:
It's possible the Qt IDE installation instructions have changed a bunch too, so if anyone has opinions about this then maybe they should check it too.

PR #275, opened 2015-10-13
Updating URL Regex

s-rah:
The old URL regex had a few issues which were revealed by fuzzing,
the biggest being that it accepted non-printable characters (e.g.
0x00 or 0x01) as part of the URL.

This created the scenario where a url of https://example.com/[0x00]
would be rendered as %2 (and attempting to open the link would give
a value like "https://example.com https://example.com " due to some
odd iteraction with the regex that I haven't quite worked out.

The new regex appears to work with all the iterations I have tried
and rejects non-printable characters.

Below are 2 picks, the first before the change with the old regex, and the second shot with the new regex.

The test strings were:

           ricochet.SendMessage("https://test.com/"+string(0x00)+"LOL", 5)
           ricochet.SendMessage("https://google.com", 5)
           ricochet.SendMessage("https://ع.com", 5)
           ricochet.SendMessage("https://عععععع<b>test</b>عععععععععع.com", 5)
           ricochet.SendMessage("https://عععععععععععععععع.com", 5)
           ricochet.SendMessage("https://ععع234233344ععععععععععععع.com/342?test=5&7=hello~", 5)

before
after

### DIFF
--- a/src/ui/LinkedText.cpp
+++ b/src/ui/LinkedText.cpp
@@ -41,7 +41,7 @@ LinkedText::LinkedText(QObject *parent)
     : QObject(parent)
 {
     // Select things that look like URLs of some kind and allow QUrl::fromUserInput to validate them
-    linkRegex = QRegularExpression(QStringLiteral("([a-z]{3,9}:|www\\.)([^\\s,.);!>]|[,.);!>](?!\\s|$))+"), QRegularExpression::CaseInsensitiveOption);
+    linkRegex = QRegularExpression(QStringLiteral("([a-z]{3,9}:|www\\.)([\\p{L}\\p{N}\\-\\._~:=&/?#]+)"), QRegularExpression::CaseInsensitiveOption);
 
     allowedSchemes << QStringLiteral("http")
                    << QStringLiteral("https")


src/ui/LinkedText.cpp
```diff
@@ -41,7 +41,7 @@ LinkedText::LinkedText(QObject *parent)
     : QObject(parent)
 {
     // Select things that look like URLs of some kind and allow QUrl::fromUserInput to validate them
-    linkRegex = QRegularExpression(QStringLiteral("([a-z]{3,9}:|www\\.)([^\\s,.);!>]|[,.);!>](?!\\s|$))+"), QRegularExpression::CaseInsensitiveOption);
+    linkRegex = QRegularExpression(QStringLiteral("([a-z]{3,9}:|www\\.)([\\p{L}\\p{N}\\-\\._~:=&/?#]+)"), QRegularExpression::CaseInsensitiveOption);
```
special: Hmm, we are losing a (useful?) trick here.

This section would capture all non-whitespace characters, excluding the characters `,.);!>` _when they are at the end of the URL_. The idea is that if you type `I use http://ricochet.im/. It's neat.`, that period doesn't become part of the link.

Is there a way we can duplicate that effect here, by excluding those characters from the first class and adding them to a second one with the negative forward lookahead?

special:
Also, nice catch - thanks for taking the time to fuzz this.

s-rah:
Closing in favor of new PR #302

PR #291, opened 2015-10-28
Prevent bidi phishing

s-rah:
Adding this is a separate PR, hence the duplicate url-regex commit.

Basically, this change is to prevent phishing attacks utilizing the unicode control characters to affect the rendering of the text so while the user sees https://ricochet.im the link actually goes to http://evil.com  - it's a fairly low level risk due to the fact a user has to add the person as a contact, and ultimately copy and paste the link etc. 
 
Sample of how the old code worked:
csyhhcxwoaa6jld png large

And the new code:

screenshot - 102815 - 18 25 57

### DIFF
--- a/src/ui/LinkedText.cpp
+++ b/src/ui/LinkedText.cpp
@@ -41,7 +41,7 @@ LinkedText::LinkedText(QObject *parent)
     : QObject(parent)
 {
     // Select things that look like URLs of some kind and allow QUrl::fromUserInput to validate them
-    linkRegex = QRegularExpression(QStringLiteral("([a-z]{3,9}:|www\\.)([^\\s,.);!>]|[,.);!>](?!\\s|$))+"), QRegularExpression::CaseInsensitiveOption);
+    linkRegex = QRegularExpression(QStringLiteral("([a-z]{3,9}:|www\\.)([\\p{L}\\p{N}\\-\\._~:=&/?#]+)"), QRegularExpression::CaseInsensitiveOption);
 
     allowedSchemes << QStringLiteral("http")
                    << QStringLiteral("https")
--- a/src/ui/LinkedText.cpp
+++ b/src/ui/LinkedText.cpp
@@ -41,7 +41,7 @@ LinkedText::LinkedText(QObject *parent)
     : QObject(parent)
 {
     // Select things that look like URLs of some kind and allow QUrl::fromUserInput to validate them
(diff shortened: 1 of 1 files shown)


s-rah:
Realized this morning while thinking about this that my original solution was completely off. Updated with a much better solution which forces the actual url within the anchor text.

src/ui/LinkedText.cpp
```diff
@@ -64,7 +64,7 @@ QString LinkedText::parsed(const QString &input)
 
         if (start > p)
             re.append(input.mid(p, start - p).toHtmlEscaped().replace(QLatin1Char('\n'), QStringLiteral("<br/>")));
-        re.append(QStringLiteral("<a href=\"%1\">%2</a>").arg(QString::fromLatin1(url.toEncoded()).toHtmlEscaped()).arg(match.capturedRef().toString().toHtmlEscaped()));
+        re.append(QStringLiteral(" <a href=\"%1\">&#8234;%2</a>").arg(QString::fromLatin1(url.toEncoded()).toHtmlEscaped()).arg(match.capturedRef().toString().toHtmlEscaped()));
```
special: What if the URL is in the middle of a block of right-to-left script? I tried a URL in the middle of arabic text, and with this patch the URL somehow appeared to the left of all of the text, instead of its correct position.

Is that character added at the beginning a space? If so, why?

Also, could you add a comment explaining what this is and why it's necessary?

I found http://unicode.org/reports/tr36/#Bidirectional_Text_Spoofing - interesting reading. Wow, unicode is crazy.

s-rah: Can you provide the sample text for the right-to-left scenario? I will take a look to see what's going on there.

That space shouldn't be there, looks like miscommited somewhere, will fix that. 

And yep. Will comment.

special: I just took a random sentence from wikipedia and typed a URL somewhere in it:

`هي لغة برمج http://www.google.com/ ة تستخدم...`

The result is at least substantially different. I don't know enough to say if it's worse.

s-rah:
Updated the diff, which fixes the inline rtl issue by adding a pop directional formatting blob at the end...testing with all relevant text strings, seems to work as expected now e.g. the link is embedded in the text as written. I'll take a look at the URL pull request tomorrow RE: not catching at the end. If this one looks good I'll cancel these 2 PR's and merge them into one since I also noticed '%' was missing from the regex.

special:
Thanks. The bidi parts are looking great.

s-rah:
Closing in favor of new PR #302

PR #294, opened 2015-10-31
RFC: Initial Draft for Group Chat Design

s-rah:
Spent some time this week thinking about group chat. Please find an initial request for comments draft.R

### DIFF
--- a/doc/multi-party-messaging.md
+++ b/doc/multi-party-messaging.md
@@ -1,0 +1,118 @@
+# Multi Party Messaging
+
+Author: Sarah Jamie Lewis
+Created: 2015-10-30
+Status: Draft
+
+
+## Overview 
+
+There has been some demand for multi-party (or group) messaging, where a single
+client can send and receive responses from multiple users in the context of 
+a single chat. This document outlines a design for the Ricochet protocol to
+implement multi-party messaging (MPM).
+
+
+## Design
+
+Alice is a Ricochet user who wishes to initiate a group chat with her contacts,
+Bob and Carol.
+
+1. Alice sets up a new hidden service HS and publishes an `openchannel` request to
+Bob and Alice with a multi-party context which contains the ricochet id for
+HS. 
+
+2. Once Bob and Carol receive the open channel message from Alice they attempt to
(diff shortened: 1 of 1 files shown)


burdges:
Why open a new HS?  Afaik it doesn't buy any security and maybe leaks info to the network.

Just a start on an alternative suggestion :  Sort out the crypto story with an eye towards this.  Users create a group chat that runs inside their existing chats by tagging messages.  Messages are simply forwarded amongst users in the pattern in which they were added.  Messages reference the hash of previous messages either of that user or an anti-chain in the directed message graph to minimize the risk of malicious users manipulating the conversation.  Ricochet id's for forwarded messages are derived by hashing with the contact id of the forwarding user, so that (a) if you've never met the person before then you do not learn their ricochet id, but (b) if you know them then you recognize them.  This design preserves lots of deniability, but makes it easy to impersonate someone.  We maybe do not need so much deniability as Tor probably does not give it anyways, but I'm a deniability addict. We 'd need to mix in something from your pairwise state with your real contacts to make it impossible to impersonate someone you actually know, or maybe just say hi to them on their real HS.

It's worth chatting with @infinity0 about this sort of thing since he's thought about it lots.

burdges:
Actually, your design maybe leaks less to the HSDirs than forwarding designs do, just some HS they've never seen before, much better than a specific groups of HSs all seeing similar traffic simultaneously.

infinity0:
I haven't thought about how to hide metadata. But yes, doing group chat well is a complex topic and it's worth studying the theoretical problems involved before proposing a hasty design that repeats historical inadequacies. (No current group messaging applications solve those problems well, and that includes Signal.)

ZimTester:
In a sad world with lack of mpOTR and stuck its development, what about look at **np1sec** ([library](https://github.com/equalitie/np1sec), [draft](https://learn.equalit.ie/wiki/Np1sec)) now instead? From one of initial mpOTR contributors, eQualitie.
Also there are another interesting implementations like in OMEMO protocol (but XMPP only) and in some messengers like Signal (formerly TextSecure).

PR #297, opened 2015-11-04
General Additions and Grammar Improvements

TheJawaEyes:


special:
Thanks!

At some point, I'd like to revise the README to be more in line with the newer language used on https://ricochet.im, and add more technical links to it.

PR #313, opened 2015-12-23
Sanitize nicknames more carefully

special:
> QML makes the questionable decision of parsing all Text items as
rich text if they appear to like HTML. This HTML parser is, fortunately,
nothing like a browser - it supports only simple visual formatting.
Still, in the worst case it can trigger network requests.

> We were being too lax with sanitizing variables before display, if they
weren't directly from an external source. In one case, this could lead
to deanonymization if a user reviews and accepts a contact request
with a nickname containing an  tag.

These commits harden against this problem in three ways: by blocking any possible network requests, ensuring nicknames are sanitized every time they're displayed, and by blocking a few more characters in incoming contact requests.

Nicknames in Ricochet are always set by the local user, so this wasn't directly a vulnerability. However, inbound contact requests can suggest a nickname, which is displayed (safely) and used if the user accepts the request without changing it. If that request was accepted despite the nickname, it would be possible for a remote user to trigger network activity.

### DIFF
--- a/src/protocol/ContactRequestChannel.cpp
+++ b/src/protocol/ContactRequestChannel.cpp
@@ -85,12 +85,17 @@ static bool isAcceptableNickname(const QString &input)
     if (input.size() > Data::ContactRequest::NicknameMaxCharacters)
         return false;
 
+    /* Although nicknames should always be escaped before being displayed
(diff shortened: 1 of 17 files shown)


rburchell:
:+1:

s-rah:
Awesome! :+1:

PR #314, opened 2015-12-23
Clear internal socket state after failed connections

### LINKED ISSUE #295: Ricochet does not show online contacts
*The other open issues have very technical titles, so I'm sorry if this is a dup of something.*

I've noticed that oftentimes Ricochet will not show an online contact when I know they are online. The only way to fix it is to quit the app and restart. This can be annoying since it means losing the contexts of all the conversations you were having. And of course it's an annoying issue in its own right.

gabedwrds:
On recent builds, Ricochet was failing to reconnect to contacts after a first failed attempt, because the connection to the SOCKS proxy was stuck in a failed state. Fix this by closing the TorSocket after a failure, making Qt reset the internal SOCKS connection.

Fixes #295

### DIFF
--- a/src/tor/TorSocket.cpp
+++ b/src/tor/TorSocket.cpp
@@ -143,6 +143,10 @@ void TorSocket::connectToHost(const QHostAddress &address, quint16 port, OpenMod
 
 void TorSocket::onFailed()
 {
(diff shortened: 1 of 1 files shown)


special:
Thanks!

PR #320, opened 2016-01-02
Add a man page source

kb100:
man page compiles with

`ronn -w -r ricochet.1.ronn`

Requires ruby-ronn package

Still needed: add compiling and installing man page to build process. (How?)

### DIFF
--- a/doc/ricochet.1.ronn
+++ b/doc/ricochet.1.ronn
@@ -1,0 +1,54 @@
+ricochet(1) -- An anonymous peer-to-peer instant messenger
+=============================================
+
+## SYNOPSIS
+
+`ricochet` [<PROFILE_DIR>]
+
+## DESCRIPTION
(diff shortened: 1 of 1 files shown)


special:
How difficult is it to build a manpage without bringing in non-standard-unix-build dependencies? I'd rather we don't need ruby.

pixelrust:
I just learnt this and I have zero experience on this. That said, it seems "groff syntax" (or whatever that's called) must be used:
http://www.tldp.org/HOWTO/Man-Page/q3.html

Here's @kb100 's content in groff syntax (not 100% complete):
https://github.com/pixelrust/ricochet/blob/master/doc/ricochet.1

`man ./ricochet.1` renders directly the file, almost correctly.

No pull requests from me yet, because I have no idea of the path where to commit the file.

PR #321, opened 2016-01-05
minijail amd64 seccomp policy: add getrandom, gettid and readlinkat

lgeek:
These syscalls are used when Ricochet starts up on my system.


special:
Merged, thanks

timkuijsten:
Just curious, wasn't the lack of getrandom a security issue?

special:
@timkuijsten no; getrandom would've failed when running under seccomp, and we would use other sources of entropy. getrandom is never used as the sole source of entropy.

Ricochet and Tor will both refuse to run if they lack sufficient entropy.

timkuijsten:
@special thanks for clearing up!

PR #344, opened 2016-02-19
Limit nick length and whitespaces in ContactUser and UserIdentity

### LINKED ISSUE #337: Disallow newlines in contact renames
This is a small UX bug which affects certain elements when a given contact is renamed and the new name includes new lines. Low impact, as the user has to take this action themselves - should be easy to clean up.

(We might also want to consider limiting the length of this field...and has some relation to #327)

image

dmS0Zq:
Fixes #337.

Max nick length of 24 was chosen arbitrarily.

It seemed necessary to make changes in both ContactUser.cpp and UserIdentity.cpp even though only the changes in ContactUser.cpp seemed to have any affect.

### DIFF
--- a/src/core/ContactUser.cpp
+++ b/src/core/ContactUser.cpp
@@ -288,7 +288,9 @@ QString ContactUser::nickname() const
 
 void ContactUser::setNickname(const QString &nickname)
 {
-    m_settings->write("nickname", nickname);
+    QString newNick = nickname.simplified();
+    newNick.truncate(24);
+    m_settings->write("nickname", newNick);
 }
 
 QString ContactUser::hostname() const
--- a/src/core/UserIdentity.cpp
+++ b/src/core/UserIdentity.cpp
@@ -128,7 +128,9 @@ QString UserIdentity::nickname() const
 
 void UserIdentity::setNickname(const QString &nick)
 {
-    m_settings->write("nickname", nick);
+    QString newNick = nick.simplified();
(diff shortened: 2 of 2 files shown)


s-rah:
I think the spirit of the change is fine - however, I think 24 is too short. As a usecase, I have many ricochet contacts who I reference by name(@twitter handle) - even taking by name only just makes the cut - "Sarah (@SarahJamieLewis)" and there are a fair few of my contacts who would be truncated by this change.

With truncation, the only time that length is an issue is in the "...is online/offline" display, truncating it at this UI point might be a nicer solution, without having to make arbitrary decisions about name length.

dmS0Zq:
Now it only makes changes to what is displayed. In this order:

* repeated whitespace characters are squashed into a single space
* leading and trailing whitespace is removed
* truncated to the first 80 characters

special:
These limits sound reasonable.

We shouldn't be changing the nickname at display-time, though. We'll end up having to copy this code to a lot of different places, and the behavior will be unexpected for some cases.

Can you apply this when setting the nickname for a contact, instead?

dmS0Zq:
Yes. So back to silently truncating, but at 80 characters instead and additonally trimming and replacing repeated whitespaces with single spaces?

special:
Sounds good.

Ideally, we should also change the UI where nicknames is set, to have the same maximum length and avoid confusing users too much.

PR #385, opened 2016-03-17
Use Tor's ADD_ONION command when available

### LINKED ISSUE #227: Use the new ADD_ONION control API when available
Tor's ADD_ONION API will enable lots of cool things for us, like private key encryption, and using system-level tor instances.

It's only available in Tor 0.2.7.x (currently alpha).

special:
For Tor >= 0.2.7, start using the ADD_ONION control command instead of SETCONF to configure hidden services, and store the private keys in our configuration file instead of as separate files on the filesystem.

Older versions of Tor remain supported, and the behavior will not change. There is an edge case where a profile is created with a new version of Tor and later downgraded - this is not supported. Don't do that.

Existing profiles will copy their keys into the configuration and stop using the files on disk, but those files will not be removed yet to help avoid any risk of data loss. They will be removed later, likely when we enable profile encryption.

Fixes #227

### DIFF
--- a/src/tor/AddOnionCommand.cpp
+++ b/src/tor/AddOnionCommand.cpp
@@ -1,0 +1,105 @@
+/* Ricochet - https://ricochet.im/
+ * Copyright (C) 2016, John Brooks <john.brooks@dereferenced.net>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *    * Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ *
+ *    * Redistributions in binary form must reproduce the above
+ *      copyright notice, this list of conditions and the following disclaimer
+ *      in the documentation and/or other materials provided with the
+ *      distribution.
+ *
+ *    * Neither the names of the copyright owners nor the names of its
+ *      contributors may be used to endorse or promote products derived from
+ *      this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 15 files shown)


ioerror:
What kind of tests do you want to see here to merge this code?

src/tor/AddOnionCommand.cpp
```diff
@@ -0,0 +1,105 @@
...
+    QByteArray out("ADD_ONION");
+
+    if (m_service->privateKey().isLoaded()) {
+        out += " RSA1024:";
+        out += m_service->privateKey().encodedPrivateKey(CryptoKey::DER).toBase64();
+    } else {
+        out += " NEW:RSA1024";
+    }
+
+    foreach (const HiddenService::Target &target, m_service->targets()) {
```
rburchell: not something you have to change now, but in the future you may want to look into range-for (and qAsConst), it avoids copies of data & leads to smaller binary size.

src/tor/AddOnionCommand.cpp
```diff
@@ -0,0 +1,105 @@
...
+        out += " Port=";
+        out += QByteArray::number(target.servicePort);
+        out += ",";
+        out += target.targetAddress.toString().toLatin1();
```
rburchell: is this really supposed to be latin1?

src/tor/AddOnionCommand.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    out.append("\r\n");
+    return out;
+}
+
+void AddOnionCommand::onReply(int statusCode, const QByteArray &data)
+{
+    TorControlCommand::onReply(statusCode, data);
+    if (statusCode != 250) {
```
rburchell: would be nice to have a constant rather than a magic number

rburchell:
:+1:

ioerror:
Works for me on SubgraphOS now. Thanks!

PR #386, opened 2016-03-17
Allow configuring external Tor in the environment

special:
```
    If the TOR_CONTROL_PORT environment variable is present, use that
    external instance of Tor instead of launching a bundled one.

    TOR_CONTROL_HOST and TOR_CONTROL_PASSWD are also available for the
    control hostname and password respectively. These variables match
    the names used by Tor Browser.

    There is no need to configure the SOCKS address; Ricochet will discover
    it automatically.
```

Also, avoid making destructive changes to a tor instance we didn't start, by tracking whether we've taken ownership or not.

### DIFF
--- a/src/tor/TorControl.cpp
+++ b/src/tor/TorControl.cpp
@@ -72,6 +72,7 @@ public:
     TorControl::Status status;
     TorControl::TorStatus torStatus;
     QVariantMap bootstrapStatus;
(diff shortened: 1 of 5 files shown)


rburchell:
:+1:

ioerror:
Works on SubgraphOS now. Thanks!

PR #391, opened 2016-03-22
Make chat window combined by default

### LINKED ISSUE #355: Make single-window the default
Myself and others I've spoken to all seem to agree the single-window view is much better from a UI/UX standpoint than the current multi-window default.

dmS0Zq:
Fixes #355

Added function to `main.cpp` where other default settings can be loaded.


special:
Thanks!

Issue #403, opened 2016-04-13
Hyperlinking breaks for some links.


rburchell:
Steps to reproduce:

* Send the string "https://lh5.googleusercontent.com/-tXQ0JeICMDA/VwK1lU6GIlI/AAAAAAAAASU/Vo0Ew4YONfAhrBhhrt6eSF8Q8-UQ6b7bw/w392-h472-no/04.04.16%2B-%2B1" (without quotes)
* Attempt to click & open the link. It won't work.
* Copy the text, paste into browser. It works.

Issue #372, opened 2016-02-29
URL hyperlinking problems with null byte


deanjerkovich:
if you send the following message, it will break string parsing:

http://www.test.com/?a=%00

resulting in a message with just '%2' being delivered, but ultimately linking to test.com

s-rah:
Yeah, there is a bigger issue here with handling unicode in the regex. PR is: #302

PR #404, opened 2016-04-14
Fix parsing of links containing escape sequences

special:
When formatting links to be displayed in messages, using repeated calls
to QString::arg() will interpret '%0' and similar in the URL as a
placeholder, resulting in the link not being properly formatted.

Depending on the value, this can either result in a broken URL when
copied to the clipboard (#403), or the URL being displayed with a label
of just '%2' (#372).

This cannot be used to mislabel links, and there is no printf-style
format vulnerability with QString::arg. There is no security impact.


rburchell:
:+1:

PR #419, opened 2016-05-20
add some osx caveats to BUILDING doc

kaizensoze:
Add caveat to get around openssl includes having been removed in El
Capitan.
Add some step by step instructions including explicitly using qt5 qmake
since brew qt5 is keg only due to potential qt4/5 conflicts.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -82,6 +82,15 @@ Normally, configuration will be stored in a `config.ricochet` folder, in the sam
 
 The `packaging/osx/release_osx.sh` script demonstrates how to build a redistributable app bundle.
 
+Since the openssl header files were removed in El Capitan, have qmake use the openssl that comes with brew (see the OPENSSLDIR var below).
+
+Steps:
+```
+brew install protobuf qt5 tor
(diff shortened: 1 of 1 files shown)


9Yg1rxeSeha90ZU1:
I approve this change. I authored it :)

special:
Oof! We should never have to edit the generated makefiles.

One better (manual) workaround is to add `OPENSSLDIR=/usr/local/opt/openssl/` to the qmake command. Documenting that is one option.

I'd be more excited to approve a change to the .pro file to automatically use `/usr/local/opt/openssl/` on mac when it exists and `OPENSSLDIR` isn't explicitly set. If you look at doing that, remember to change `tst_cryptokey.pro` too.

kaizensoze:
Would it be ok to replace the makefile edit suggestion in the doc with the OPENSSLDIR var one for this PR and then take a look at the .pro file in a separate PR at a later time?

special:
Yes, absolutely

kaizensoze:
This should be good to go with the OPENSSLDIR change.

special:
Thanks!

PR #436, opened 2016-06-21
passing parseError to accurately get any errors;

zethon:


special:
Good catch. I squashed out the NUL file that snuck into your commit and merged as 4633c9cae9e97a3d208e676f8f42d41a927063d4. Thanks.

PR #444, opened 2016-08-09
add support for OpenSSL 1.1

satta:
OpenSSL 1.1 introduces some API changes (https://wiki.openssl.org/index.php/1.1_API_Changes) which affect building of Ricochet. This PR adds some extra code to enable compiling with both the old and the new API. *Please review carefully before merging!*

### DIFF
--- a/src/utils/CryptoKey.cpp
+++ b/src/utils/CryptoKey.cpp
@@ -129,9 +129,13 @@ bool CryptoKey::loadFromFile(const QString &path, KeyType type, KeyFormat format
 
 bool CryptoKey::isPrivate() const
 {
-    const BIGNUM *p, *q;
-    RSA_get0_factors(d->key, &p, &q);
-    return isLoaded() && (p != 0);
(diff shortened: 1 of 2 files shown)


src/utils/CryptoKey.cpp
```diff
@@ -119,12 +129,14 @@ bool CryptoKey::loadFromFile(const QString &path, KeyType type, KeyFormat format
 
 bool CryptoKey::isPrivate() const
 {
-    return isLoaded() && d->key->p != 0;
+    const BIGNUM *p, *q;
+    RSA_get0_factors(d->key, &p, &q);
```
special: This will crash in the `!isLoaded()` case

special:
Otherwise, I think this looks ok. We'll want to try building with LibreSSL and make sure that still works too. Thanks!

satta:
I think addressed the issue you were raising. Thanks for the feedback!

Issue #343, opened 2016-02-19
OS X: Minimize window using ⌘M


rebpim:
I use `⌘M` frequently to minimize but it doesn't seem to work with Ricochet.. would it be possible to add in `⌘M` support?

CR0CKER:
Or maybe leave the app running when closing the window, as is the default behavior of native OS X applications?

tnull:
+1 to both of these. Maybe this should be called 'make ricochet behave like a Mac app'.

PR #469, opened 2016-10-23
keep app running even if window closed [on OS X]

kaizensoze:
The default behavior of native OS X apps is to leave the app running
even on closing the last remaining window. Since the
quitOnLastWindowClosed property of QApplication is true by default, we
can remove the line of code that explicitly quits the app on the main
window not being visible.

reference: #343

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -69,6 +69,10 @@ int main(int argc, char *argv[])
     a.setWindowIcon(QIcon(QStringLiteral(":/icons/ricochet.svg")));
 #endif
 
+#if defined(Q_OS_MAC)
+    a.setQuitOnLastWindowClosed(false);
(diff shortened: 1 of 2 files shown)


special:
This is incomplete: after closing the window, there is no way to open it again without quitting. Generally with Mac applications, clicking the dock will bring the window back.

It's also important that the unread badge and icon bounce continue to work when the window is closed. One option might be to hide() instead of close() the window, and figure out how to unhide when the dock icon is activated.

kaizensoze:
Actually this pull request is unnecessary. There are plenty of single window OS X apps that quit on x'ing out the window (App Store, Calculator, Blackmagic Disk Speed Test, etc.). I think I'll just close this and look into another issue.

PR #472, opened 2016-11-02
Fix cross-compiled Windows build

special:
Minor fixes to allow cross-compilation for Windows.

### DIFF
--- a/ricochet.pro
+++ b/ricochet.pro
@@ -130,7 +130,7 @@ win32 {
     }
 
     # required by openssl
-    LIBS += -lUser32 -lGdi32 -ladvapi32
+    LIBS += -luser32 -lgdi32 -ladvapi32
 }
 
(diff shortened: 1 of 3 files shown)


ricochet.pro
```diff
@@ -263,7 +263,7 @@ TRANSLATIONS += \
     translation/ricochet_zh_HK.ts
 
 isEmpty(QMAKE_LRELEASE) {
-    win32:QMAKE_LRELEASE = $$[QT_INSTALL_BINS]\lrelease.exe
+    contains(QMAKE_HOST.os,Windows):QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease.exe
```
special: @rburchell is there a nicer way to figure out the name of `lrelease`, or to figure out where there's a `.exe` suffix for the host (not target!) platform?

rburchell: Not sure offhand

Issue #60, opened 2014-07-17
Dead keys and other input methods don't work with linux static builds


obvio171:
Must be an issue with Qt, but can't type characters like 'á', 'ã', 'ç', etc.

special:
@rburchell points out that we should make sure the `src/plugins/platforminputcontexts/compose` Qt plugin is built in. Adding something to `QTPLUGIN` in the .pro file might help here.

special:
That hint was correct, but we ideally want ibus too. To compile those plugins into the linux-static build, ricochet.pro needs to contain:

```
QTPLUGIN.platforminputcontexts = composeplatforminputcontextplugin ibusplatforminputcontextplugin
```

The ibus plugin requires qtdbus, but this already seems to be available in the build-time environment and isn't creating any additional runtime dependencies.

Using only compose would probably only partially solve the problem, and some desktops (e.g. Gnome) set `QT_IM_MODULE=ibus` in the environment, which breaks if we only have compose.

PR #477, opened 2016-11-04
Deploy linux input plugins and remove unneeded plugins

special:
For Linux, we need to include the ibus and compose input plugins for
some input methods to work (e.g. dead keys). See issue #60.

Also, we don't need playlists, mediaservice, imageformats, or printing.
Remove those plugins from static builds. Unfortunately, this doesn't
remove them from deployment with windeployqt or macdeployqt.

### DIFF
--- a/ricochet.pro
+++ b/ricochet.pro
@@ -133,6 +133,14 @@ win32 {
     LIBS += -luser32 -lgdi32 -ladvapi32
 }
 
+# Exclude unneeded plugins from static builds
+QTPLUGIN.playlistformats = -
+QTPLUGIN.imageformats = -
+QTPLUGIN.printsupport = -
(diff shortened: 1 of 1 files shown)


ricochet.pro
```diff
@@ -133,6 +133,14 @@ win32 {
     LIBS += -luser32 -lgdi32 -ladvapi32
 }
 
+# Exclude unneeded plugins from static builds
+QTPLUGIN.playlistformats = -
+QTPLUGIN.imageformats = -
+QTPLUGIN.printsupport = -
+QTPLUGIN.mediaservice = -
+# Include Linux input plugins
```
rburchell: It might be nice to note why this is needed.

Issue #184, opened 2015-04-27
building on debian jessie - translations are not compiled with lrelease


esme-does-code:
following the steps in BUILDING.md, qmake gives you a load of errors about missing files (translation_*.qm).

workaround: build like this
$ lrelease ricochet.pro
$ qmake
$ make

i would send you a patch, but i'm not qmake's biggest fan and haven't really used it much before

related to: #57

system: debian jessie. dependencies installed as listed in the help files.

special:
These errors are actually pretty harmless. The files it's talking about are generated when lrelease is called at build-time, and they will be used when the resource compiler actually executes. From what I understand, the only effect here is the annoying messages, and that it may not generate the right Makefile rules to rebuild the resource if .ts files are changed.

One potential workaround ([source](https://stackoverflow.com/questions/4366898/how-to-specify-auto-generated-resource-files-to-qmake)) is to generate them at qmake-time with `system()` too. @rburchell, does that seem reasonable to you?

myfreeweb:
Just built on FreeBSD 11-CURRENT, had to do this as well. This should be at least documented in `BUILDING.md`

special:
@rburchell ping for an opinion on this problem

PowerKiller:
I hit it too, lrelease worked for me.

PowerKiller:
#269 and #184 is what I hit every time I try to build on Debian/Ubuntu. :(

rburchell:
QMAKE_EXTRA_COMPILERS can probably be used to add an extra (automatic) target that will run lrelease at build time.

(sorry @special, for some reason I never saw the original messages at me)

special:
Fixed by #503. Thanks.

PR #503, opened 2016-12-30
Various small improvements to the build system

rburchell:
Should help address #184.

### DIFF
--- a/ricochet.pro
+++ b/ricochet.pro
@@ -247,43 +247,41 @@ lupdate_only {
 
 # Translations
 TRANSLATIONS += \
-    translation/ricochet_en.ts \
-    translation/ricochet_it.ts \
-    translation/ricochet_es.ts \
-    translation/ricochet_da.ts \
-    translation/ricochet_pl.ts \
-    translation/ricochet_pt_BR.ts \
-    translation/ricochet_de.ts \
-    translation/ricochet_bg.ts \
-    translation/ricochet_cs.ts \
-    translation/ricochet_fi.ts \
-    translation/ricochet_fr.ts \
-    translation/ricochet_ru.ts \
-    translation/ricochet_uk.ts \
-    translation/ricochet_tr.ts \
-    translation/ricochet_nl_NL.ts \
-    translation/ricochet_fil_PH.ts \
-    translation/ricochet_sv.ts \
-    translation/ricochet_he.ts \
-    translation/ricochet_sl.ts \
-    translation/ricochet_zh.ts \
-    translation/ricochet_et_EE.ts \
-    translation/ricochet_it_IT.ts \
-    translation/ricochet_nb.ts \
-    translation/ricochet_pt_PT.ts \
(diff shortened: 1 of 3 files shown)


rburchell:
Sorry this is a little meatier than you'd hope for; I've been sitting on some of these changes in some form for a while.

ricochet.pro
```diff
@@ -106,21 +106,26 @@ win32-msvc2008|win32-msvc2010 {
...
 INCLUDEPATH += src
 
+win32|mac {
+    # For mac, this is necessary because homebrew does not link openssl .pc to
+    # /usr/local/lib/pkgconfig (presumably because it used to be a system
+    # package).
+    #
+    # Unfortunately, it is no longer really a system package, and we really
+    # need to know where it is.
+    isEmpty(OPENSSLDIR): error(You must pass OPENSSLDIR=path/to/openssl to qmake on this platform)
```
special: This does unfortunately mean that macs with a valid `pkg-config libcrypto` won't use it. But testing `packagesExists(libcrypto)` would be insufficient because macOS' old 0.9.8 seems to leave behind some broken pkgconfig files (and because of the no-pkg-config problem).

So, the workaround is annoying and I'm ok with this for the current state of things.

ricochet.pro
```diff
@@ -242,43 +247,38 @@ lupdate_only {
...
+    ricochet_nb \
+    ricochet_pt_PT \
+    ricochet_sq \
+    ricochet_zh_HK \
+    ricochet_ja
+
+contains(QMAKE_HOST.os,Windows):QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease.exe
+else:QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease
+for (translation, TRANSLATIONS) {
+    system($$QMAKE_LRELEASE translation/$${translation}.ts -qm translation/$${translation}.qm)
```
special: This is running several times during qmake -- once per pass, I assume. Is there a way to run it only once?

Out of scope, but I wonder if a qmake expert could be tricked into writing up a `translations.prf` (or pri) that takes a variable listing .ts files, `lrelease`s them all, and generates a resource without needing to write a separate .qrc. Seems possible?

rburchell: I'm not quite sure what you mean by that. I only see it run once (although once for each language file) -- how are you building?

Yes, such a thing could be possible, perhaps (with the caveat that I'm not sure how to solve the rcc dependency issue), although I'm not sure it's of much benefit when done here. I guess ideally Qt would gain such functionality that (most of) initTranslation would become a library call (along with built-in resourcing & automated lrelease as you mention).

special: A `git clean` build is showing three runs through the whole list of translation files. I think two of them are skipped for subsequent runs (along with e.g. the hardening tests).

rburchell:
Multiple runs should be fixed now. Fingers crossed.

Issue #122, opened 2014-10-22
OS X packaging should make the installation process more clear


idevk:
Cannot create directory: /Volumes/Ricochet/config.ricochet

special:
I think there are actually a few changes worth making here, because this seems to be a confusing process for many people:

- [ ] Add background graphics to the .dmg, and a link to Applications, so it's easier to figure out
- [ ] We should also detect when running inside the .dmg, and either:
 -  [ ] Show a _useful_ error message, or
 -  [ ] Offer to install (or extract?) right there, and then re-execute automatically from the correct location

tfeb:
For what it's worth, on OSX 10.12 it still fails to run from the DMG with a very unencouraging error message:

```
Cannot create directory: /private/var/folders/78/msz6jjzs6f304f2n632z3f040000gp/T/AppTranslocation/A2DB0E26-72DE-4892-A3D4-3143DE9764B6/d/config.ricochet
```

I can see why it's hard to make it run from the DMG, but it should give a useful message about it.  That likely needs some horrid special case in the code ('am I on `/Volumes/*` and is it writable'), which I can see is a pain.

I would *not* like it to try to automatically install itself: software which wants to copy itself onto my machine is always a bit worrying (not to mention it would choose the wrong location).

holmesworcester:
I just had two friends hit this problem too, and I'm really excited about Ricochet since I've been thinking about p2p chat a ton for my project https://zbay.app

So I got inspired and decided to take this on :)  Here's the PR: #607 

And here's what it looks like on my machine: 

<img width="770" alt="image" src="https://user-images.githubusercontent.com/213678/87596671-87a45080-c6be-11ea-9d3b-316eae30a473.png">

PR #504, opened 2016-12-30
main: Check for a path that looks like an OS X bundle, and give a better error

rburchell:
Saying "Can't create path /private/..." is really not very
user-friendly.

Contributes to #122.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -184,7 +184,14 @@ static bool initSettings(SettingsFile *settings, QLockFile **lockFile, QString &
 
     QDir dir(configPath);
     if (!dir.exists() && !dir.mkpath(QStringLiteral("."))) {
-        errorMessage = QStringLiteral("Cannot create directory: %1").arg(dir.path());
+#ifdef Q_OS_MAC
+        QString absolutePath = QFileInfo(QLatin1String(".")).absoluteFilePath();
+        if (absolutePath.startsWith(QLatin1String("/private")) ||
+            absolutePath.startsWith(QLatin1String("/Volumes")))
+            errorMessage = QStringLiteral("Please move Ricochet outside of the DMG before attempting to run it.");
+        else
+#endif
+            errorMessage = QStringLiteral("Cannot create directory: %1").arg(dir.path());
         return false;
     }
 


rburchell:
NB: I have not tested this beyond compilation. I haven't got an environment set up to create DMG.

src/main.cpp
```diff
@@ -184,7 +184,14 @@ static bool initSettings(SettingsFile *settings, QLockFile **lockFile, QString &
 
     QDir dir(configPath);
     if (!dir.exists() && !dir.mkpath(QStringLiteral("."))) {
-        errorMessage = QStringLiteral("Cannot create directory: %1").arg(dir.path());
+#ifdef Q_OS_MAC
+        QString absolutePath = QFileInfo(QLatin1String(".")).absoluteFilePath();
+        if (absolutePath.startsWith(QLatin1String("/private")) ||
+            absolutePath.startsWith(QLatin1String("/Volumes")))
```
special: This has some false positives. Any external disk is going to be under `/Volumes`. The impact of that is not huge -- just a confusing error message -- but, hmm. Just noting.

src/main.cpp
```diff
@@ -184,7 +184,14 @@ static bool initSettings(SettingsFile *settings, QLockFile **lockFile, QString &
...
+            errorMessage = QStringLiteral("Please move Ricochet outside of the DMG before attempting to run it.");
```
special: We could probably nitpick and debate the language for awhile. I don't know if this error means anything more to most people who run into this problem than the current one does. I don't know if it should be translated.

I'm more inclined to not be picky about it, though.

src/main.cpp
```diff
@@ -184,7 +184,14 @@ static bool initSettings(SettingsFile *settings, QLockFile **lockFile, QString &
 
     QDir dir(configPath);
     if (!dir.exists() && !dir.mkpath(QStringLiteral("."))) {
-        errorMessage = QStringLiteral("Cannot create directory: %1").arg(dir.path());
+#ifdef Q_OS_MAC
+        QString absolutePath = QFileInfo(QLatin1String(".")).absoluteFilePath();
```
rburchell: Oh, true enough!

src/main.cpp
rburchell: I agree, absolutely. I can't find a "nice" way to do this concisely unfortunately.

src/main.cpp
rburchell: I'll try tinker with the wording a bit; I think it's a little on the abrupt side myself.

rburchell:
I have no plans to work on this, as I no longer use a mac on a daily basis to test on.

PR #513, opened 2017-01-16
Add .travis-ci.yml

special:
Based off of the rules @kolodziej wrote long ago in #261 (with my apologies for not working more on it then!), these seem to be working travis rules for a Linux and OSX build.

This currently does nothing other than test compilation. The trusty Qt version is very old (5.2), so QML likely won't be very functional. I couldn't find a usable PPA offering any reasonable Qt versions.

Still, seems kinda useful. This could be tied into Coverity later also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,44 @@
+language: cpp
+
+os:
+  - linux
+  - osx
+
+env:
+  - CONFIG=Debug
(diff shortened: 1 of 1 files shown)


special:
Test builds are at https://travis-ci.org/special/ricochet

special:
Approved out of band by several people.

Issue #345, opened 2016-02-19
Handling large numbers of messages


s-rah:
Ricochet will store messages at any rate until ultimately running out of memory. Locally on my small system I can get up to ~1M messages before ricochet consistently crashes.

While certain edge cases are unavoidable, we should determine a strategy for Ricochet when there are a large number of messages in memory. 

A few possible strategies:
- Ricochet could do nothing. "This is a user problem - they should get more RAM..."
- Ricochet could limit the number of messages it accepts from each contact - and refuse newer messages.
- Ricochet could cache old messages to a file .
- Ricochet could delete old messages in favor of newer messages (once they have been read, or even before if there are enough messages)
- A combination of the above.

Likely I think the best solution is to provide a default (say 1000 messages) and to delete older messages regardless of read state - users can increase or decrease this default in the settings (most of the text messaging apps I've used have operated this way)

Relatedly, there are also UX issues when the number of messages reaches a certain threshold: The scrollbar completely vanishes making it difficult to determine where in the thread you are.

mva1985:
As an example the surespot messaging app allows 1000 messages per conversation.  Once reached the oldest are deleted.

mateli:
Some of us like to have our message history so deleting old messages wouldn't be a good solution. Denying messages for flood protection would be the way to go. 

Also messages should be persisted to disk using a very stable database format. Then limiting messages in ram should be done, messages should be loaded from disk on demand.

wfr:
A simple suggestion:
https://github.com/ricochet-im/ricochet/compare/master...wfr:issue-345-limit-history?expand=1
This branch adds an option that limits the history size. Old messages are automatically pruned. The default limit is 1000 messages per conversation, as per mva1985's suggestion.

rburchell:
@wfr Nice. I suggest opening a pull request for it :)

I'm not @special, so take my suggestions with a grain of salt, but I'm not totally sure if exposing it in the UI like that is great.

What do you think about time-based expiry rather than purely number-of-messages? Something like a combo box like this (excuse the shitty ASCII art):

    Delete messages after [ 7 days ]
                          [ 30 days ]
                          [ 30 days ]

[...]

PR #516, opened 2017-01-16
Limit conversation history to 1000 entries (#345)

wfr:
From issue #345:

@s-rah 
>Ricochet will store messages at any rate until ultimately running out of memory. Locally on my small system I can get up to ~1M messages before ricochet consistently crashes.

This patch prunes old messages in the chat window to a hardcoded limit.

### DIFF
--- a/src/core/ConversationModel.cpp
+++ b/src/core/ConversationModel.cpp
@@ -117,6 +117,7 @@ void ConversationModel::sendMessage(const QString &text)
     beginInsertRows(QModelIndex(), 0, 0);
     messages.prepend(message);
     endInsertRows();
+    prune();
 }
 
 void ConversationModel::sendQueuedMessages()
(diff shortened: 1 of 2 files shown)


src/core/ConversationModel.cpp
```diff
@@ -316,3 +318,13 @@ int ConversationModel::indexOfIdentifier(MessageId identifier, bool isOutgoing)
...
 }
 
+void ConversationModel::prune()
+{
+    const int history_limit = 1000;
+    while(messages.size() > history_limit)
+    {
+        beginRemoveRows(QModelIndex(), messages.size()-1, messages.size()-1);
+        messages.removeLast();
+        endRemoveRows();
```
special: Fun Qt fact: a _lot_ of things are recalculated, laid out, possibly rendered, and otherwise inefficiently processed on model change signals.

In this case it doesn't matter, because it will never end up in a position to remove more than one row. Otherwise, I'd insist on changing it to a single `beginRemoveRows(QModelIndex(), history_limit, messages.size()-1)`.

I'm happy either way here.

special:
Thanks for taking the time to submit this!

wfr:
Thanks for your feedback.
I improved the efficiency, just in case the function is ever called with more than N+1 stored messages.

Issue #278, opened 2015-10-15
crash on removing a pending contact (Linux64, 1.1.1)


Yawning:
Installed the package via AUR but I can also reproduce this by pulling down git master and doing a debug build.
1. Add a contact but do not accept it.
2. Remove the pending contact, either via `Right Click->Remove` or by `Preferences->Contacts->Remove`.

My binary crashes with:

```
src/core/ContactUser.cpp:331:38: runtime error: member call on null pointer of type 'struct QObject'
```

ContactUser.cpp:ContactUser::deleteContact():

```
    if (m_contactRequest) {
        qDebug() << "Cancelling request associated with contact to be deleted";
        m_contactRequest->cancel();
        m_contactRequest->deleteLater(); // <- HERE
    }
```

OutgoingContactRequest.cpp:OutgoingContactRequest::removeRequest():

```
    /* Clear the request settings */
    m_settings->undefine();
    emit removed(); // <- NULLs out m_contactRequest.
```

Not sure if removing the `deleteLater()` call is sufficient here, but this should be enough details for you to fix it. cbf sending a patch.

s-rah:
Looks like this was fixed by #266 - can't reproduce in 1.1.2

pokulo:
This Bug might not be hit, since its race condition via event calling. 
The bad call sequence looks like this:

[ContactUser::deleteContact  # m_contactRequest->cancel();](https://github.com/ricochet-im/ricochet/blob/master/src/core/ContactUser.cpp#L329)
[ OutgoingContactRequest::cancel  # removeRequest();](https://github.com/ricochet-im/ricochet/blob/master/src/core/OutgoingContactRequest.cpp#L195)
[OutgoingContactRequest::removeRequest  # emit removed();](https://github.com/ricochet-im/ricochet/blob/master/src/core/OutgoingContactRequest.cpp#L169)
The race:
[ContactUser::requestRemoved  # m_contactRequest = 0;](https://github.com/ricochet-im/ricochet/blob/master/src/core/ContactUser.cpp#L358)
before (bad!) / after (ok)
[ContactUser::deleteContact  # m_contactRequest->deleteLater();](https://github.com/ricochet-im/ricochet/blob/master/src/core/ContactUser.cpp#L330)

**And its still present in current master!** Not fixed by https://github.com/ricochet-im/ricochet/pull/266

wfr:
Reproducible in current master:

[...]

PR #522, opened 2017-01-31
Avoid race condition (ricochet-im/ricochet#278)

wfr:
between ContactUser::deleteContact()
and ContactUser::requestRemoved()

### DIFF
--- a/src/core/ContactUser.cpp
+++ b/src/core/ContactUser.cpp
@@ -328,9 +328,6 @@ void ContactUser::deleteContact()
         qDebug() << "Cancelling request associated with contact to be deleted";
         m_contactRequest->cancel();
     }
-    if (m_contactRequest) {
-        m_contactRequest->deleteLater();
-    }
 
     emit contactDeleted(this);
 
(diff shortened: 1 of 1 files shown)


rburchell:
Isn't this essentially dead code? `m_contactRequest->cancel()` calls `OutgoingContactRequest::removeRequest`, which unconditionally emits `removed`, which should then call into `ContactUser::requestRemoved`, which does a deleteLater and assigns to 0.

Maybe dropping a `Q_ASSERT(!m_contactRequest);` and removing the deleteLater (since it should always be dead AFAICT) would be a clearer solution?

src/core/ContactUser.cpp
```diff
@@ -327,6 +327,8 @@ void ContactUser::deleteContact()
     if (m_contactRequest) {
         qDebug() << "Cancelling request associated with contact to be deleted";
         m_contactRequest->cancel();
+    }
+    if (m_contactRequest) {
         m_contactRequest->deleteLater();
```
pokulo: just delete the two new lines and "m_contactRequest->deleteLater();" from here, since it will be called via cancel() anyways.

pokulo: applied my proposed changes to my code 1.1.4 release, and could finally delete that pending contact.

wfr:
Thanks for your feedback.
I'm opening a new pull request because it's silly to have 3 commits for 2 changed lines.

PR #527, opened 2017-02-17
Fix LibreSSL compatibility

hasufell:
This is just a quick fix, there might be more sophisticated ways to check for the availability of those functions without relying on version-specific ifdefs.

Also: __I do not know if the resulting codepath is safe__. Please review not just this diff, but the resulting codepath.


special:
I just merged #529 with exactly the same fix, before I noticed this one. That makes me feel better about the solution being correct :)

Thanks!

PR #529, opened 2017-02-25
Fix build with LibreSSL

Sp1l:
- Check LIBRESSL_VERSION_NUMBER in addition to OpenSSL > 1.1

LibreSSL forked from OpenSSL 1.0.1f and does not have all features
from 1.1.0.


special:
Merged (without those .orig files) as 79c79a0398de1557c7c572327247eaa31e3a7942. Thanks!

PR #554, opened 2017-08-13
Bounty source Salt and Bounty campaign button

increasingawareness:
If @special thinks we should still do this.


rburchell:
Thanks for your contribution. For future reference, it looks like you might want to look into how to get to a clean git state (read up on git fetch & git reset, or git pull --rebase).

This isn't a big problem for now though, no need to do anything on here as it can be fixed at our end as is.

I won't merge this myself though, as while I contribute here I don't make the decisions, so to speak.

PR #563, opened 2017-11-23
Fixes in ricochet.pro: used $PREFIX, avoid adding /usr to INCLUDEPATH

yurivict:
This is upstreamed from the FreeBSD port.


yurivict:
Ping. Could you please merge this trivial patch?

PR #585, opened 2018-07-16
WIP create v3 onion service (Fixes #575)

### LINKED ISSUE #575: Hidden Services v3
Ricochet currently uses classic Tor hidden service addresses. Next generation hidden services fix security and design flaws found in the original or legacy hidden services.

Onion v3 functionality is now enabled in the stable release version of Tor.

Ricochet should support this new address type.

* [Specification: Next-Generation Hidden Services in Tor](https://gitweb.torproject.org/torspec.git/tree/proposals/224-rend-spec-ng.txt)
* [YouTube: DEF CON 25 - Roger Dingledine - Next Generation Tor Onion Services](https://www.youtube.com/watch?v=Di7qAVidy1Y)  
Roger Dingledine actually mentions Ricochet in this talk :)

kradan:
according to `config/tor/state` this commit has been tested with
TorVersion Tor 0.3.3.9 (git-ca1a436fa8e53a32)

In the UI the dialog to add contacts still shows a short id (ricochet:dkska2iuk3byot7g) and I expected it be longer, so maybe this change was not enough to create a v3 onion service yet.

I am waiting here for interested people to test if connections work (ricochet is not always on).

### DIFF
--- a/src/tor/TorControl.cpp
+++ b/src/tor/TorControl.cpp
@@ -522,6 +522,8 @@ void TorControlPrivate::publishServices()
                                  .arg(tit->targetAddress.toString())
                                  .arg(tit->targetPort);
                 torConfig.append(qMakePair(QByteArray("HiddenServicePort"), target.toLatin1()));
+                QString version = QString::fromLatin1("3");
+                torConfig.append(qMakePair(QByteArray("HiddenServiceVersion"), version.toLatin1()));
             }
 
             QObject::connect(command, &SetConfCommand::setConfSucceeded, service, &HiddenService::servicePublished);


duffert (commented):
Nice try, but this will not work. Even though you set the correct HiddenServiceVersion, the internal structure of Ricochet needs to be updated in order to support HiddenService version 3. We need to be able to load (and store) version 3 private and public keys, as well as to be able to handle the different cryptographic algorithems associated with v2 and v3 HiddenService identifiers.

kradan:
Thanks for the explanation, I expected something like that :)
As Ricochet is work in progress (without much progress at the moment though), I updated the original ticket.

ashleyconnor:
V3 support [was recently added to stem](https://stem.torproject.org/api/descriptor/hidden_service.html#stem.descriptor.hidden_service.HiddenServiceDescriptorV3) so this might be possible now.

Issue #561, opened 2017-11-18
Installation error on OSX (High Sierra)


d3prof3t:
I'm facing an issue with the installation process. Please refer to the link for screenshot of the same [here](https://share.riseup.net/#5_OBselnFc6cOdfB97ee4w
)

d3prof3t:
I've found a solution to this using `brew cask`. If anybody requires then refer [here](http://macappstore.org/ricochet/) for instructions.

rburchell:
glad you found a workaround. the issue itself is a duplicate of #122.

in future, please upload things to Github directly or transcribe them (drag and drop, or clipboard) - having the content directly in the issue tracker is a lot more future-proof.

Issue #510, opened 2017-01-08
UX: Package DMG with an "Applications" folder alias and a background (OSX)


jpt:
I had a friend ask about the error when double-clicking the Ricochet application from within mounted DMG. 

Something not unlike this.

(no reason for this particular image, first google image result)

![install](http://www.luxrender.net/wiki/images/d/dc/OSX_luxrender_install.png) 

An even better solution that I've seen on modern Mac apps is double-click opens, and then asks if you want to move the app to the Applications folder, but that may be a lot more work.

special:
Definitely agree. See #122 for more discussion around this. I believe it's easy to include a background like this, if we had one.

Issue #491, opened 2016-12-18
Error on installation "Cannot create directory ..."


robotme:
Cannot create directory: /private/var/folders/yd/x6hyxwdn5pd76ss_znft2dwm0000gn/T/AppTranslocation/51963695-B778-49FC-8179-BB180E53F964/d/config.ricochet

Are other people experiencing this? 

MacOS 10.12.2

Sorry if this is the wrong place to ask. I don't know where else to ask troubleshooting questions for most open source applications found on GitHub.

special:
We really need to get this one fixed.

You need to drag Ricochet out of the .dmg to a location of your choice (such as the Applications folder) before running it. This error is trying, very badly, to say that you can't run from inside a disk image.

See #122

Issue #112, opened 2014-10-07
Serve download+instructions HTML page from RicoID's onion address


hsribei:
Currently I have to separately give out my ID to a person, then instructions on how to download Ricochet and add me.

Since we're already running a hidden service, a greatly simplified way of getting people to contact you (and, thus, to grow Ricochet's install base more quickly) would be to just send them your URL (http://ricoid.onion) or put it on your profiles elsewhere on the Internet.

That would serve a static page with detailed instructions (screenshots, screencast-like animated gifs, text) on how to download and add you for a newcomer, along with a download button that links to the official release (can't trust the contact to serve the binaries).

If that's too much (worries of DoS, perhaps?), it could just redirect to an official Ricochet page with the id as a query parameter. That page generates instructions based on the query parameter, kind of like this page does (beautiful idea!): http://jekyllbootstrap.com/usage/jekyll-quick-start.html

PR #607, opened 2020-07-15
Adds background image and Applications shortcut to macOS disk image

holmesworcester:
Many users have reported confusion with the `Cannot create directory: /Volumes/Ricochet/config.ricochet` error they see when neglecting to drag Ricochet.app out of the disk image. See #112, #561, #510, #491, etc. 

Ricochet is still an amazing proof of what's possible for private, peer-to-peer chat and it's a shame that so many Mac users will hit this bug, so I got inspired and fixed it. I tested Ricochet with two friends and **both** experienced this issue and assumed Ricochet simply didn't work. (One friend is a junior software engineer, and the other is very technical. It seems like that this bug is preventing a *lot* of users from trying Ricochet!)

This PR modifies `release_osx.sh` to add a shortcut to the user's Applications folder and a background image reminding the user to drag the file, using the [create-dmg](
https://github.com/create-dmg/create-dmg) script and following the default UX of [electron builder](https://github.com/electron-userland/electron-builder/). It uses a [weird-and-tricky-to-create](https://stackoverflow.com/questions/11199926/create-dmg-with-retina-background-support/11204769#11204769) background image ([this one](https://github.com/electron-userland/electron-builder/blob/e1a7ec9f3f7dde69b99dc8c809aac688803ee9a9/packages/dmg-builder/templates/background.tiff)) from the [electron builder](https://github.com/electron-userland/electron-builder/) project. Electron-builder is available under an MIT license, so the use of this image should not present any problems. Mac users will see the following when they download Ricochet:

image

My own peer-to-peer messaging project [Zbay](https://zbay.app) uses the same default electron-builder DMG layout as the one above, and user typically drag the app to their Applications folder. It would be possible to add branding here, but that seems like overkill.

I'm a junior developer, so apologies for the funky commit history, but it works and seems sound. I wasn't sure if [this commit](https://github.com/ricochet-im/ricochet/pull/607/commits/104449ed8d1d7a488e32079b7b35b0225c3ed35e) made sense, but it was necessary for me to get the `release_osx.sh` working locally so I figured it might be helpful to others.


holmesworcester:
Also, it may be worth testing distributing Ricochet as a .zip file. I *think* some browsers unzip it by default to the downloads folder, at which point it would be impossible for users to make the mistake this PR addresses. Many heavily user-tested apps (Spotify, I think?) use this approach, but it's a bit less common. This might be worth trying too and I'd be up for trying it if the PR would be accepted.

cypherbits:
Hello @holmesworcester , Ricochet IM is an abandoned project. This repository is not updated.

There was an attempt to renew it on 
https://github.com/blueprint-freespeech/ricochet-refresh 
https://ricochetrefresh.net/
but it is already abandonware too.

We look forward to Cwitch.im, based on Ricochet but written in Golang.

holmesworcester:
@cypherbits bummer! let me know if the above packaging work would be helpful to Cwtch!

m-simonelli:
Hi @holmesworcester,
We at blueprint are actively working on ricochet refresh, so feel free to make a pull request to our build repo over at [https://github.com/blueprint-freespeech/ricochet-build](https://github.com/blueprint-freespeech/ricochet-build)

